PR #197, opened 2014-08-19
Added EMFM redirect to wiki

alistairuk:
This is to allow a simple emfcamp.org URL to be read on air. Just a redirect to the wiki under the same domain. No visual impact on the site.


marksteward:
Closing this for now - I think radio could have its own memorable domain next time, and if not, it's easy enough to add. Thanks for the PR anyway!

Issue #105, opened 2012-08-01
Put token_only in the DB


marksteward:
token_only should be an attribute of TicketType, not some arbitrary code. There's no need to do any complex processing here.

PR #216, opened 2015-08-22
Store ticket token in db

### LINKED ISSUE #108: Store ticket token in ticket
So we can track origins from the DB

SamLR:
This closes #108 although adds another use of token_only for clean up in #105.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ clean:
 db:
 	$(ENV) SETTINGS_FILE=$(SETTINGS) python ./utils.py createdb
 
-data: db tickets bankaccounts
+data: db tickets bankaccounts tokens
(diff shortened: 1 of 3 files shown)


marksteward:
Tried merging this, but it only works in development because sqlite is overly permissive. Check out the schema for TicketToken - the token is a keyword that enables potentially multiple ticket types to be discounted/made available. Tokens also expire, and a keyword can be reused. So the ticket should be tied to an id, not the keyword itself.

marksteward:
Maybe token should be renamed to keyword or something similar which makes it clear it's just free text.

marksteward:
(note, the branch can't be safely deleted - please open a new PR)

PR #219, opened 2015-08-24
CFP email and tidier templates

### LINKED ISSUE #183: Send an email when a proposal is submitted

SamLR:
Several small fixes

- Fixes #183 (email captures title, description, duration, attendees etc) for submissions to CFP.
- Improves dev logging of emails (now lists to, from, subject and body) which is required for testing email templates
- Moves email templates into their own sub-directory of templates


marksteward:
Nice - thanks :D

PR #220, opened 2015-08-29
Update vagrantfile, use debian 8, use postgres 9.4

marekventur:
The Vagrantfile looked a bit derelict, I've update it to make setting up a dev environment easier. 

* Uses debian 8 (the url is a minimal debian image a colleague of mine has created)
* Uses Postgres 9.4 - because debian 8 doesn't support 9.1 out of the box and why not
* Uses Vagrant config type 2
* Makes flask bind to all interfaces instead of just localhost - I guess the production box will be firewalled off and it makes port forwarding in vagrant a lot easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Requirements
 Starting
 ========
 ```
-sudo apt-get install python-dev python-virtualenv libxml2-dev libxslt-dev libffi-dev postgresql-server-dev-9.1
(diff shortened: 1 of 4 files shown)


marksteward:
Is this missing a `provision.sh`?

marekventur:
Doh! That's very possible. I'll check tonight when I'm at home.

marekventur:
Added.

marksteward:
Thanks for this - looks good!

Issue #234, opened 2015-10-06
Fix highlighting of invalid inputs


marksteward:
We don't seem to be getting highlights on invalid forms fields now (although we do get `help-inline`). Make sure this is visible.

SamLR:
Closing as we've now fixed this with #249 and #254

PR #249, opened 2015-11-15
help-inline is not supported anymore, using form-group and help-block instead

marekventur:
See #234 and http://code.divshot.com/bootstrap3_upgrader/
 
I've also replaced ```help-inline``` with ```help-block``` everywhere else, but I haven't managed to test them (how can you trigger an error on a checkbox? How can I trigger an error on the phone number field?)

error


marksteward:
To test, add something like:


```
    def validate_volunteer(form, field):
        raise ValidationError('test')
```

to FullTicketForm.

PR #266, opened 2015-12-05
Replace TicketAttrib transfer with class

SamLR:
Rather than use key/value pairs in the TicketAttrib table use a specific
table to log transfers. This commit also improves the user information
on transfers.


russss:
Seems legit.

PR #292, opened 2015-12-30
Db based feature flags

### LINKED ISSUE #276: Move TICKET_SALES state to DB
So we don't have to restart the app every time, causing 500s

Would be nice to include CFP too.

SamLR:
Closes #276

### DIFF
--- a/models/feature_flag.py
+++ b/models/feature_flag.py
@@ -1,0 +1,15 @@
+from main import db, cache
+
+class FeatureFlag(db.Model):
+    __tablename__ = 'feature_flags'
+    name = db.Column(db.String, primary_key=True)
+    enabled = db.Column(db.Boolean, nullable=False)
+
+    def __init__(self, name, enabled=False):
+        self.name = name
+        self.enabled = enabled
+
+    @classmethod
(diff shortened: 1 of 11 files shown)


models/feature_flag.py
```diff
@@ -0,0 +1,15 @@
...
+    __tablename__ = 'feature_flags'
+    name = db.Column(db.String, primary_key=True)
+    enabled = db.Column(db.Boolean, nullable=False)
+
+    def __init__(self, name, enabled=False):
+        self.name = name
+        self.enabled = enabled
+
+    @classmethod
+    @cache.cached(timeout=30, key_prefix='feature_flags_get')
```
russss: I think you need to use [cache.memoize](https://pythonhosted.org/Flask-Cache/#flask.ext.cache.Cache.memoize) here to take the function arguments into account.

This also means you can do [cache.delete_memoized](https://pythonhosted.org/Flask-Cache/#flask.ext.cache.Cache.delete_memoized) in the admin interface to invalidate immediately.

marksteward:
:grin:

Issue #69, opened 2012-06-17
Deal with in-progress cancellation


marksteward:
Split out from #22:

- We need to request that GoCardless cancel the payment, and deal with any complications that causes.
- All cancellations need to be kept around anyway, in case the money's sent in the end.

marksteward:
We get an email from GoCardless if an uncharged payment is left too long. Low priority, as this only happens when a browsing session is terminated (perhaps after 500).

marksteward:
We can't cancel a GoCardless payment once it's been sent to the bank (can_be_cancelled=False), but cancellation works otherwise.

For cancelled BankTransactions, we resurrect the payment if it comes through, and email the user. So I don't think we need to change anything.

PR #293, opened 2015-12-31
Cancellable gocardless

### LINKED ISSUE #282: Fix admin payment cancellation
Currently 404s

SamLR:
Closes #282 and allows manual closing of payments for #69


russss:
I think it would probably be worth checking if the gocardless payment can be cancelled (it's the "can_be_cancelled" attribute) and not trying to actually cancel it if it can't.

marksteward:
:dancer:

PR #344, opened 2016-04-10
Add DB migrations using flask-migrate

russss:


marksteward:
According to the Alembic docs, "Autogenerate can’t currently, but will eventually detect: Some free-standing constraint additions and removals, like CHECK, PRIMARY KEY - these are not fully implemented." It might also not deal with defaults.

I think regardless of whether we use Alembic, we should have a script that builds the latest schema and then diffs it against the current DB.

PR #370, opened 2016-05-15
Make user lookup by email case insensitive

Jonty:
If a user signs up on a device that happens to capitalise the first letter of the email alias, then later attempts to login on a device that _doesn't_ happen to capitalise it, we'd like them to be able to login.

We do this rather than lowercasing all email addresses because email aliases can be case sensitive.

Appears to work as expected in a VM.


marksteward:
:grin:

PR #375, opened 2016-05-23
Tweak copy before accepting messages

Jonty:
@marksteward can you quickly sanity check this?


Jonty:
@russss confirms it looks ok. Merging.

PR #393, opened 2016-06-20
Make the "we have deferred you" email clearer

Jonty:
Given everyone was so confused by the last email we want to make sure everybody knows their talk has not been rejected.


marksteward:
:railway_car:

PR #630, opened 2018-03-11
Add rough availability percentages to "not on sale"

Jonty:
This PR is open for discussion and poking, I don't think this initial implementation should ship.

Points:
* Should we make this automatic
* How should we communicate it


vetraskr:
From the perspective of keeping vendors happy, we have to be careful that our wording doesn't give the impression that sales aren't strong. I'm not sure how that could be phrased though.

If a programmatic approach is taken, we'd still have to apply some sort of fuzzing/rounding to the result. Manual changes would be a simpler approach, and if we keep them fairly broad won't require updating _too_ often (at least once the demand slackens somewhat).

SamLR:
I don't know if that's really a worry now? Given the whole reason for showing this text at all is that we've sold out all available tickets (so far) in less than a minute...

russss:
This probably needs transposing into an issue if we want to do it next time. I don't think it's a concern any more this year.

Issue #438, opened 2016-07-27
Allow users to create an account somehow


marksteward:
For volunteers and favouriting things from the schedule. Only enable it shortly before the event, to encourage people to transfer tickets.

davidc:
partly done (with USER_SIGNUPS feature flag), but a question before I send a pull request: do you want it to also automatically log them in or just automatically send them a login link?

marksteward:
Log them in, I think. Once you have an account, you can always request a login link, so there's no point making them do the little dance.

marksteward:
Thinking more, it's probably a bad idea to create an account with nothing associated. We should do the login dance, same as when admins create an account.

marksteward:
I think for now I'm also going to require a time-limited token from an admin, to prevent random spam.

marksteward:
Closed by #636

PR #636, opened 2018-03-12
Admin ticket reservations and user registration

### LINKED ISSUE #602: Work out a clean way for people to pay for tickets outside the usual flow
As usual, we need to let people buy tickets outside of the usual rounds, and without forcing a payment type on them.

Previously we just created a new ticket type and added a token, but that now involves a ProductViewProduct (and new ProductGroup if we want to limit them).

We could make this easier, or create a way to pre-populate a basket and start them at the /ticket/pay step. Alternatively, allow payments that are "new" to change provider.

marksteward:
Allow admins to:
 - give a user a link to let them create an account (/admin/user/signup)
 - reserve tickets for a user once they have an account (/admin/tickets/reserve)

Allow users to:
 - sign up with an admin-provided link (/signup)
 - recover their reserved tickets into their basket (/tickets/reserved)

Also tidy the login and "my account" pages and fix allow_promo.

Security: this changes verify_login_code into a slightly more generic verify_hmac_msg.

Fixes #602, #438


apps/users.py
```diff
@@ -106,6 +111,71 @@ def logout():
...
+    name = StringField('Name', [Required()])
+    allow_promo = BooleanField('Send me occasional emails about future EMF events')
+    signup = SubmitField('Sign up')
+
+    def validate_email(form, field):
+        if User.does_user_exist(field.data):
+            field.was_duplicate = True
+
+            msg = Markup(
+                render_template_string('Account already exists. '
```
SamLR: will these strings be auto-concatenated?

apps/users.py
```diff
@@ -106,6 +111,71 @@ def logout():
...
+
+@users.route("/signup", methods=['GET', 'POST'])
+def signup():
+    if not request.args.get('code'):
+        abort(404)
+
+    uid = verify_signup_code(app.config['SECRET_KEY'], time.time(), request.args.get('code'))
+    if uid is None:
+        flash("Your signup link was invalid. Please note that they expire after 6 hours.")
+        return redirect(url_for('.account'))
```
SamLR: Shouldn't this redirect to a login page rather than the account page?

apps/users.py
marksteward: yep

apps/users.py
marksteward: it will, indirectly. I'll make that more explicit

PR #660, opened 2018-05-19
Some code of conduct finesse

russss:
This is based on some reckons I had this evening about using "we" to try and make it clearer that the code of conduct is a collective thing. A few other tweaks as well.
image

I'm happy to make more changes and I'm not averse to expanding the CoC a little.

CC @tef and @drcable if that even works?

### DIFF
--- a/templates/code-of-conduct.html
+++ b/templates/code-of-conduct.html
@@ -7,29 +7,31 @@
 for everyone.</p>
 
 <p>
-   Sadly, technology events have a notorious reputation for harassment. We've set these ground
+   Sadly, technology events have a notorious reputation for harassment. We set these ground
    rules to make it abundantly clear that this kind of behaviour is not acceptable, so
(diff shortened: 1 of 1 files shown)


marksteward:
Not sure if this is useful, but as nobody else has commented...

- "sexualised imagery in public spaces" probably belongs in a separate item, unless I've misunderstood the intent
- I like the collective obligation idea, but 'we' seems like it refers to the organisers until you get to "We will not engage in aggression or elitism", which seems out of place (maybe because it sounds like a creed, or maybe because it's followed by a 'we' that shifts focus back to the organisers). I'd be tempted to revert this line unless we go wholesale for a collective voice, which I suspect would also make it more of an "accord with" than "abide by".

All the other changes look good to me.

russss:
I think I'll leave this and reconsider it after this event.

PR #686, opened 2018-06-23
Handle the awful python header wrapping crash.

Jonty:
I am not happy with this, but I also do not care about it very much. We could do with better logging somewhere for when we're failing to send emails.


russss:
We did #687 for this instead.

Issue #657, opened 2018-05-15
GC payments can be failed after being paid out


marksteward:
```
Apr 25 14:35:44 maxwell pipenv[16991]: 2018-04-25 14:35:44,743 apps.payments.gocardless INFO     None         Webhook resource type: payments, action: paid_out, cause: payment_paid_out, origin: gocardless
Apr 25 14:35:44 maxwell pipenv[16991]: 2018-04-25 14:35:44,745 root         INFO     None         Received paid_out action for gcid ..., payment ...
...
Apr 30 09:08:26 maxwell pipenv[13538]: 2018-04-30 09:08:26,011 apps.payments.gocardless INFO     None         Webhook resource type: payments, action: failed, cause: insufficient_funds, origin: bank
Apr 30 09:08:26 maxwell pipenv[13538]: 2018-04-30 09:08:26,015 apps.payments.gocardless INFO     None         Received failed action for gcid ..., payment ...
Apr 30 09:08:27 maxwell pipenv[13538]: 2018-04-30 09:08:27,938 apps.payments.gocardless ERROR    None         Current payment state is paid (should be inprogress), ignoring
```

marksteward:
Fixed by 189d3b1. To sanity check, I've done this:

```
In [19]: for p in ps:
    ...: p.gc = gocardless_client.payments.get(p.gcid)
    ...: p.expected_state = {'cancelled': 'cancelled', 'failed': 'failed', 'confirmed': 'paid', 'paid_out': 'paid', 'pending_submission': 'inprogress', 'submitted': 'inprogress', 'charged_back': 'refunded'}[p.gc.status]

In [22]: [(p, p.state, p.gc.status) for p in ps if p.state != p.expected_state]
Out[22]:
[(<GoCardlessPayment ...>, 'cancelled', 'failed'),
 (<GoCardlessPayment ...>, 'paid', 'failed'),
 (<GoCardlessPayment ...>, 'cancelled', 'failed'),
 (<GoCardlessPayment ...>, 'refunded', 'paid_out'),
 (<GoCardlessPayment ...>, 'cancelled', 'failed'),
 (<GoCardlessPayment ...>, 'cancelled', 'failed')]
```

The cancelled ones are fine - we don't cancel the GC payment when unreserving the tickets. Likewise the refunded. So I've just updated the state the remaining one that failed.

PR #703, opened 2018-07-18
Failure is an option

marksteward:
To address #657.


russss:
What's the rationale for adding the `late-failed` state rather than just using `failed`? I'm a bit concerned it might be easy to miss that state (as it's not very common).

(Perhaps it might be better to have an array column of previous states if we need visibility into the exact state changes?)

marksteward:
That's just there in case we need to treat them differently (e.g. sending for emails). I don't think it's crucial, and we can reconstruct it from payment.versions if necessary.

russss:
I think if we want to send a different email it's probably best to pass both the previous state and the current state into the email template, rather than having a completely new state.

marksteward:
Yeah it's just an in-case. I'll remove it.

PR #756, opened 2018-08-23
Add flag to indicate whether we can send messages to a volunteer during the event

vetraskr:
I'd appreciate some feedback on this - it's the first bit of python I've done, and while I think it's technically correct, I'm uncertain whether there are any style rules I've missed.

Also, I can't actually test it. I get a 404 when going to the volunteer page, on master and in this branch.


migrations/versions/2628a398bab8_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '2628a398bab8'
+down_revision = '5a5373c0d0d0'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('volunteer', sa.Column('allow_comms_during_event', sa.Boolean(), nullable=False))
```
vetraskr: <@ms7821> the only thing I spotted is you'll need to add a server_default=expression.false() to this https://github.com/emfcamp/Website/pull/756/files#diff-8960f7762c417815fab99c7b646ea18dR19
11:45:40

<@ms7821> check out migrations/versions/716a1e3b4f86_add_cfp_accepted_only.py for an example

PR #769, opened 2018-08-26
Make sure the allow comms field is populated

jellybob:
The data was being persisted, but that wasn't being reflected on
the form.


static/js/volunteer-schedule.js
```diff
@@ -1,4 +1,5 @@
 function init_volunteer_schedule(data, all_roles, active_day) {
+   console.log("Is this doing anything at all?")
```
Jonty: Probably want to remove this!

Jonty:
LGTM other than the console.log needing removing.

PR #772, opened 2018-08-27
Volunteer UI Improvements

jellybob:
Most notably I've brought the schedule view in line with the rest of the site, and I've added the start of support for markdownified role descriptions to be displayed when looking at details.

### DIFF
--- a/static/js/volunteer-schedule.js
+++ b/static/js/volunteer-schedule.js
@@ -62,10 +62,13 @@ function init_volunteer_schedule(data, all_roles, active_day) {
 
     function make_open_modal_fn(shift) {
         return function open_modal() {
-            $('#signUp .modal-title').html('Sign up for ' + shift.role.name + ' @ ' + shift.start_time);
+            $('#modal-role').html(shift.role.name);
+            $('#modal-time').html(shift.start_time);
 
-            $('#signUp .modal-body').empty();
-            $('#signUp .modal-body').append(make_modal_details(shift));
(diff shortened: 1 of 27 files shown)


static/js/volunteer-schedule.js
```diff
@@ -74,36 +80,6 @@ function init_volunteer_schedule(data, all_roles, active_day) {
...
-            needed;
-
-        if (shift.min_needed === shift.max_needed) {
-            needed = shift.min_needed;
-        } else {
-            needed = shift.min_needed + ' - ' + shift.max_needed;
-        }
-
-        if (shift.is_user_shift) {
-            res.unshift(make_ele('p', 'You are currently signed up to this shift'));
```
Jonty: Is this functionality no longer needed?

Jonty:
LGTM, other than not being sure about the JS being removed - do we no longer need that functionality?

(Disclaimer: I've not had time to run this yet)

static/js/volunteer-schedule.js
jellybob: The same functionality is provided by the sign up buttons, which make it clearer in the list view that you're signed up for a given shift. Volunteer counts are also displayed there.

jellybob:
The same functionality is provided by the sign up buttons, which make it clearer in the list view that you're signed up for a given shift. Volunteer counts are also displayed there.

Jonty:
Merged and deployed

PR #791, opened 2018-09-01
abstract and description should not be redundant

saerdnaer:
typically they are appended by consuming clients. 

For the sake of completeness:
Frab defines '_abstract_' as  
> "One short paragraph that succinctly describes the event."

and '_description_' as 
> "A longer description of the event. Use this to give additional details that did not fit in the abstract. Both, abstract and description, will be shown on the conference website. (HTML)"


marksteward:
Thanks, merged and deployed!

PR #804, opened 2019-01-20
Dockerise the development environment

jellybob:
All the tests pass, and some clicking around hasn't surfaced anything obvious.


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3-stretch
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+      python3-dev libxml2-dev libxslt1-dev libffi-dev git glpk-utils && \
+    rm -rf /var/lib/apt/lists/* && \
+    pip3 install pipenv
+
+ADD . /app
```
russss: This isn't strictly necessary in development, as `/app` is mounted as a volume, and it'll slow the build time down as well as generate more intermediate docker images locally.

I suggest we remove this, then we can add a different Dockerfile for production when the time comes.

(Strictly it would also be [stylistically better](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) to use `COPY` rather than `ADD` here.)

russss:
Thanks! This looks like a very good start. I think I'd like to hold off on merging this until we've finished with the 2018 website and cleared everything out (after t-shirt sales end in Feb). 

Stuff like your Makefile changes may well be fine but they'll conflict with the current production configuration. It's not worth trying to make everything work with the production environment when we'll be blowing it away.

Dockerfile
jellybob: Switched that to copy only the files needed to bootstrap the virtualenv.

russss:
@SamLR just found this while trying this on a new machine:

`dev_server.py` [only listens](https://github.com/emfcamp/Website/blob/master/dev_server.py#L52-L59) on 127.0.0.1 *unless* there's a `.inside-vagrant` file present (and this file is `.gitignore`d). So it won't work on a clean install.

We should probably just make the dev server listen on 0.0.0.0 in all circumstances.

russss:
I merged this, but Github didn't apparently notice.

PR #817, opened 2019-07-29
More visible diversity statement

### LINKED ISSUE #321: Diversity statement should be more accessible
Currently it (https://www.emfcamp.org/diversity) is only reachable via a body text link on the CfP. It should at least be linked from the Code of Conduct if not as part of the footer.

SamLR:
Closes #321 

Depends on #804


russss:
I'm going to be honest here - although I wrote this originally, I'm not really sure what this diversity statement is *for*. I'm also not especially happy with the tone of it (as originally written - not due to your changes, Sam).

I think if this is intended to be a policy for our own reference, then it shouldn't be on the main website. On the other hand, if it's meant to be for (potential) attendees then I think this information should be where it's actually needed - i.e. stuff about the CfP should be on the CfP page, and stuff about tickets should be on the tickets page. Diversity should be baked in rather than tacked on as an additional page.

I think there is also scope for a general "about EMF as an organisation" page which could cover some of the general stuff. 

But I feel like the diversity statement as it stands doesn't really seem to solve an obvious problem, and we should probably just get rid of it.

SamLR:
@russss I think that makes sense. Adding an "about EMF (org)" page makes a lot of sense (also a good location to potentially link philosophy page etc.)

Agree that the statement itself should have a specific purpose. Might be worth asking Alia for thoughts?

russss:
Cool, let's repurpose #817 to try and reflect that, and I'll close this PR (which is a conflicty mess anyway).

PR #823, opened 2019-11-02
Revise Code of Conduct

russss:
This adds a bit more detail on how we will respond to CoC issues, as
well as some other miscellaneous tweaks.

### DIFF
--- a/templates/code-of-conduct.html
+++ b/templates/code-of-conduct.html
@@ -40,7 +40,11 @@
 
 <h3>Our Response</h3>
 
-<p>During our events, we aim to reach a decision on all conduct issues within 12 hours.
+<p>
+  During our events, we aim to reach a decision on all conduct issues within 12 hours.
+   Outside of events, we will initially respond within one week.
+</p>
+<p>
     All reports will be taken seriously and will be considered in confidence by our
     Conduct &amp; Accessibility team with the involvement of the Directors and
     Duty Site Manager if necessary.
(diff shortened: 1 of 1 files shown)


templates/code-of-conduct.html
```diff
@@ -3,41 +3,60 @@
...
-  All our volunteer staff will be briefed on how to handle a harassment situation.</li>
+  All our volunteer staff will be briefed on how to handle reports.</li>
   <li>Drop us an email at <a href="mailto:{{ config['DUTY_EMAIL'][1] }}">{{ config['DUTY_EMAIL'][1] }}</a>.</li>
 </ul>
-<p>If you have any comments or queries, you can get in touch with our Conduct &amp; Accessibility team at <a href="mailto:{{ config['CONDUCT_EMAIL'][1] }}">{{ config['CONDUCT_EMAIL'][1] }}</a>.</p>
 
-<p><small>This Code of Conduct is licensed under the <a href="https://creativecommons.org/licenses/by/3.0/">Creative Commons Attribution 3.0 License</a>.</small></p>
+<h3>Our Response</h3>
+
+<p>During our events, we aim to deal with all conduct issues within 12 hours.
```
Jonty: Is it worth specifying the response time when we're not running an event? (Apols for slow reply)

russss: Ideally yes, but I don't want to make promises we can't keep, so I have no idea what timeframe would be reasonable.

Jonty: Is "deal with" implying that we'll finish handling with the issue within that 12 hour period, or just that our initial response will be within 12 hours?

russss: Good point, that's not terribly clear. I changed it to "reach a decision".

Jonty: In which case we should say something like "outside of events, we aim to initially respond within one week".

I don't feel comfortable committing to a resolution time outside events, but I feel we can at least say we'll respond saying "we have received your email" within a week or so.

russss: I think that's fair, I'll add that. I'd like to be able to promise a shorter time than that but we need to make sure we'll meet it.

Issue #812, opened 2019-07-24
CFP form updates


liedra:
Some updates to the CFP form/text are necessary post-2018:

- In CFP, encourage people to do Q&A outside the tent: "Please note that in the event of timing overrun, you may not be able to have Q&A time as part of the presentation; please be prepared to take questions outside the tent after your talk is over."
- Make it obvious that there are no speaker fees paid (check box for understanding) 
- Make it obvious that ⅓ ticket subsidy (check box for understanding) and that they will definitely have a reserved ticket if talk is accepted
- Default for checkbox for needing use of laptop should be unchecked but required (this came up in 2018)

russss:
> In CFP, encourage people to do Q&A outside the tent: "Please note that in the event of timing overrun, you may not be able to have Q&A time as part of the presentation; please be prepared to take questions outside the tent after your talk is over."

This and the laptop question should happen after talk acceptance, I think, so I've moved them to issues #827 and #828 respectively so they're not blocking opening the CfP.

> Make it obvious that there are no speaker fees paid (check box for understanding)
> Make it obvious that ⅓ ticket subsidy (check box for understanding) and that they will definitely have a reserved ticket if talk is accepted

I'm concerned about the complexity of the CfP form becoming rather intimidating, so my preference is against adding more checkboxes. I've changed the wording to try and make this clearer - those changes are in #829.

russss:
I had a bit of a chat to Jonty about this a few days ago and I believe after the changes in #829 it's now in a situation where it's acceptable to both of us. 

It's not quite as pushy about the the stuff as was originally envisaged but it is clearer.

PR #829, opened 2019-11-17
CfP Wording improvements for 2020

russss:
This is an attempt to fix the main issue in #812 by changing the wording of the CfP intro like so:
image

Any other wording updates are also in scope for this PR, I have a few more tweaks which I'll commit shortly.


Jonty:
This is good, but we also need to mention the child tickets too.

I'm not so certain about the wording before the submit button as it comes across quite harshly. Will see if I can draft an update.

PR #830, opened 2019-11-18
More about the event content

### LINKED ISSUE #798: More "about the event" content
We should move some of the essential info about the event from the wiki to the main site so it's more visible, easy to read, and persists between events. Including:

* Facilities on site (toilets/showers/vendors/etc)
* Info about tickets (when they're issued, transferring them, whether you need to print them out, etc)
* When you can arrive/leave, when things start, etc.
* Accessibility
* FAQ? Or perhaps get rid of the FAQ in favour of info pages.

jellybob:
Somewhat resolves #798, although probably not completely.

### DIFF
--- a/templates/about/_nav.html
+++ b/templates/about/_nav.html
@@ -10,5 +10,6 @@
   {{item("Accessibility", "base.accessibility")}}
   {{item("Power", "base.power")}}
   {#{{item("Internet", "base.internet")}}#}
+  {{item("Childcare", "base.childcare")}}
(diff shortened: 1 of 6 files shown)


russss:
Why didn't the tests run? I have no idea.

I can spot a few typos there but looks good otherwise, so I'll just merge and fix them.

BTW GitHub is not clever enough to recognise "Somewhat resolves" and so it's going to auto-close the issue, but honestly I think we've got enough content here to start with so I think the ticket can be closed.

PR #835, opened 2019-12-04
815 improve vouchers

### LINKED ISSUE #815: Ticket Sales Voucher System
We need to allow people to buy tickets using vouchers. We will be issuing vouchers to volunteers to allow them to buy tickets outside of the public sale.

This is a sketch design document for vouchers - suggestions are welcome.

## Overall Design

Unlike the CCC system, we should ensure that *all vouchers guarantee the holder a ticket until their expiry date*. This means that the capacity logic needs to change to take unredeemed vouchers into account.

(Some?) vouchers should also give the holder an opportunity to buy friends/family tickets - perhaps one additional adult ticket and a number of U18/U12 tickets.

It's possible we might also want to issue a single voucher code for a group of people to buy tickets, so perhaps the number of adult tickets the voucher issues should be configurable.

## Implementation Thoughts

It's probably worth having a "voucher class" entity, which each voucher is a member of, to set the parameters of the voucher. The voucher class should be linked to the ProductGroup which the voucher will issue tickets from.

SamLR:
Closes #815 (mostly)

This adds a new class `Voucher` that replaces the `token` column in controlling visibility of a particular `product_view`.

It also adds a 'bulk add' form to the product view that will generate and email people individual vouchers. The [template](https://github.com/emfcamp/Website/blob/815-improve-vouchers/templates/emails/volunteer-voucher.txt) should be proofed (cc. @Jonty) before merging.

I've tested the golden path on this but there're a couple of rough edges
* Template needs proofing
* Currently if the voucher has been used the `/tickets/voucher/<xxx>` endpoint will 404, we might want to replace this with a more friendly "invalid voucher" page
* We should probably make it clear on the users' accounts page that tickets were bought via a voucher

Expected workflow:

1. Create a volunteer price tier for full tickets with a user allocation of 2 (default is 10)
2. Make sure the volunteer price tier is the only active 'full tier' (I'm assuming we don't care about limiting the number of supporter tickets here)
3. Create a product view for the tickets we want volunteers to have access to
4. use the bulk add feature to generate vouchers for volunteers

Other stuff

I replaced all incidents of 'token' (for vouchers) with 'voucher' or 'voucher_code'. There are too many tokens


models/product.py
```diff
@@ -375,41 +450,36 @@ def get_by_name(cls, name):
...
             return False
 
-        if self.token:
-            if user_token == self.token:
-                return True
+        if self.vouchers:
+            if not voucher:
+                return False
+
+            for vchr in self.vouchers:
```
russss: I think I'd rather this was a query than iterating over all vouchers, as at the moment it's going to load all vouchers in for a productview every time it gets hit.

PR #840, opened 2019-12-30
Improve homepage info on the CfP

SamLR:
This adds state info on the CfP to the home page. Ideally this should make it easier to know what's going on.

The actual text needs improving.


SamLR:
The general design needs improvement but I suspect the whole front page needs a touch up

Screen Shot 2019-12-30 at 20 06 29

PR #848, opened 2020-01-07
Poetry 1.0.0 update

jayaddison:
A couple of minor issues encountered during local setup under `poetry==1.0.0`:

1. The project `description` field in `pyproject.toml` is now required
1. The `editable` flag on git source references fails schema validation

Regarding the last item - `poetry` installs packages as editable by default, so hopefully this shouldn't negatively affect any development workflows (see https://github.com/python-poetry/poetry/issues/1120).

Also includes a lock refresh via `poetry update` at the time of writing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4,7 +4,7 @@ description = "A database migration tool for SQLAlchemy."
 name = "alembic"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-version = "1.3.1"
+version = "1.3.2"
 
 [package.dependencies]
 Mako = "*"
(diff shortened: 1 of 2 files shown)


poetry.lock
```diff
@@ -54,6 +54,12 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 version = "19.3.0"
 
+[package.extras]
```
jayaddison: I think/hope that `poetry` now renders the `package.extras` list for each package directly into the lockfile, hence these few additional entries.  Would be nice to have a changelog reference for that though.

jayaddison: Not totally certain (would have to do a bit of testing against previous versions), but it looks like this was introduced in https://github.com/python-poetry/poetry/pull/1205/commits/c29174809a0570d32686fc5689b0cd93797747da#diff-8f7a4c4761eabce6d17bef6ca760e2d7R211-R212, where the `locker` code now includes extras if present.  That was rolled into the first few `1.0.0` release candidates, and into the eventual release version.

jayaddison: Ok, yep - this definitely changed between `poetry==0.12.17` and `poetry==1.0.0a0`, and in particular it was indeed commit c29174809a0570d32686fc5689b0cd93797747da which introduced the change.  From that point on, `extras` are rendered into the lockfile and this is definitely intended.

PR #854, opened 2020-01-11
Remove gulp-util dependency

jayaddison:
[CVE-2019-10744](https://github.com/advisories/GHSA-jf85-cpcp-j695) showed up via GitHub as a vulnerability related to `lodash`, which is pulled in via [gulp-util](https://github.com/emfcamp/Website/blob/792c62080f834fb5f5864112caa257737624df84/package-lock.json#L4062).

This is only a [devDependency](https://github.com/emfcamp/Website/blob/792c62080f834fb5f5864112caa257737624df84/package.json#L41), but even so - it's easy to [replace gulp-util with minimist](https://stackoverflow.com/a/23088183) since it's only being used to parse command-line arguments; it seems worth removing the security vuln noise from github.


russss:
Thanks, that was annoying me a bit!

jayaddison:
No worries :)

PR #855, opened 2020-01-15
Provide more useful information about EMP

jellybob:
I'm not sold on that mailing list signup box, but it is at least consistent with other ones around the site.

Screenshot 2020-01-15 at 14 00 10


russss:
I don't think we're going to be announcing all EMP events through the main mailing list so I'd rather not have the signup box there at the moment. Otherwise looks good, thanks.

jellybob:
Updated to remove the EMP Events section.

PR #865, opened 2020-02-16
Code of conduct changes 2020 02 16

adamamyl:
Few updates to add from a few other good Codes of Conduct (a) discovered, or (b) mentioned by others contacting us.

### DIFF
--- a/templates/code-of-conduct.html
+++ b/templates/code-of-conduct.html
@@ -14,7 +14,7 @@
 
 <h3>The Rules</h3>
 
-<p>We do not tolerate harassment or discriminatory behaviour of any form. Specifically:</p>
+<p>We do not tolerate abusive, harassment or discriminatory behaviour of any form. Specifically:</p>
 
 <ul>
   <li>Do not engage in homophobic, biphobic, transphobic, racist, ableist, sexist, deliberate misgendering, or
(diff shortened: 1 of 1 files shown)


russss:
I am rather protective of this code of conduct - I'd really like to keep it as concise and snappy as possible, so I have a predisposition against these exhaustive lists of things. I'd like this to be something that people can easily understand, rather than a legal-esque document. 

This CoC descends from another which I was involved in writing, and we spent weeks trying to make something which was clear and easy to read. So please excuse my pedantry here.

I did have a change written about the sexualised imagery thing already, which I'm mostly happy with: "Sexualised language or imagery is not appropriate in public, or in sessions which don't deal directly with such topics"

Deliberate misgendering is already pretty clearly transphobic.

I think the "previous attendance" point is already covered by the "aggression and elitism" point, though we might be able to finesse the wording of that line a bit more.

The "Offensive comments" point is almost entirely just duplicating the first point IMO.

The "Unwelcome comments" point, I'm honestly not sure about. It's extremely broad and open to interpretation, and I feel like it covers a number of things which I don't think the code of conduct should be dealing with. I'm not really sure it's necessary and it would at any rate be quite hard to enforce.

If you're OK with it, I'll take some of the points here and push an update to the CoC myself with them.

adamamyl:
> If you're OK with it, I'll take some of the points here and push an update to the CoC myself with them.

Absolutely ;)

adamamyl:
> "Sexualised language or imagery is not appropriate in public, or in sessions which don't deal directly with such topics"

Is succinct. I think it should be clear that includes creations/maker-stuff, too.

russss:
I updated the code of conduct in 457861d - I think this is good enough for the moment.

PR #872, opened 2020-02-17
Refactor volunteer signup page to autogen dates

### LINKED ISSUE #871: Volunteer signup dates are for 2018
Our signup form has dates for 2018.

File `apps/volunteer/sign_up.py` lines 31 onwards;
```
ARRIVAL_CHOICES = [
    ("2018-08-29", "Wednesday 29 August or earlier")
] + generate_day_options("2018-08-30", "2018-09-02")
DEPARTURE_CHOICES = generate_day_options("2018-08-31", "2018-09-03") + [
    ("2018-09-04", "Tuesday 4 September or later")
```

Carding this so we don't forget it.

idnorton:
Volunteer arrival and departure dates were hard coded, this change replaces this with two config settings to determine how many arrival days or departure days the event should have.

Config items:

  ARRIVAL_DAYS = 2
  DEPARTURE_DAYS = 2

Fixes #871

### DIFF
--- a/apps/volunteer/sign_up.py
+++ b/apps/volunteer/sign_up.py
@@ -75,7 +75,6 @@ def sign_up():
     form.departure.choices = generate_departure_options()
     form.departure.data = event_end().strftime("%F")
 
-    # On sign up give user 'volunteer' permission (+ managers etc.)
     if current_user.is_authenticated and VolunteerUser.get_for_user(current_user):
         return redirect(url_for(".account"))
 
@@ -94,6 +93,7 @@ def sign_up():
         new_volunteer = update_volunteer_from_form(new_volunteer, form)
         db.session.add(new_volunteer)
 
+        # On sign up give user 'volunteer' permission (+ managers etc.)
         current_user.grant_permission("volunteer:user")
 
         db.session.commit()
--- a/apps/volunteer/sign_up.py
+++ b/apps/volunteer/sign_up.py
@@ -8,6 +8,8 @@ from flask import (
     Markup,
     render_template_string,
 )
+
+from datetime import timedelta
 from flask_login import current_user
 from wtforms import StringField, SelectField, SubmitField, BooleanField
 from wtforms.validators import Required, Email, ValidationError, Optional
@@ -15,6 +17,7 @@ from wtforms.validators import Required, Email, ValidationError, Optional
 from pendulum import parse, period
 
 from main import db
+from models import event_start, event_end
 from models.volunteer import Volunteer as VolunteerUser
 from models.user import User
 
@@ -23,28 +26,13 @@ from ..common.forms import Form
 from ..common import create_current_user, feature_flag
 
 
-def generate_day_options(start, stop):
-    days = period(parse(start), parse(stop)).range("days", 1)
-    return list([(d.strftime("%Y-%m-%d"), d.strftime("%A %-d %B")) for d in days])
-
-
-ARRIVAL_CHOICES = [
-    ("2018-08-29", "Wednesday 29 August or earlier")
-] + generate_day_options("2018-08-30", "2018-09-02")
-DEPARTURE_CHOICES = generate_day_options("2018-08-31", "2018-09-03") + [
-    ("2018-09-04", "Tuesday 4 September or later")
-]
-
-
 class VolunteerSignUpForm(Form):
     nickname = StringField("Name", [Required()])
     volunteer_email = StringField("Email", [Email(), Required()])
     over_18 = BooleanField("I'm at least 18 years old")
     volunteer_phone = StringField("Phone", [Required()])
-    arrival = SelectField("Arrival", choices=ARRIVAL_CHOICES, default="2018-08-31")
-    departure = SelectField(
-        "Departure", choices=DEPARTURE_CHOICES, default="2018-09-03"
-    )
+    arrival = SelectField("Arrival")
+    departure = SelectField("Departure")
     allow_comms = BooleanField(
         "May we send you messages during the event?", [Optional()]
     )
@@ -82,8 +70,12 @@ def update_volunteer_from_form(volunteer, form):
 @feature_flag("VOLUNTEERS_SIGNUP")
 def sign_up():
     form = VolunteerSignUpForm()
-    # On sign up give user 'volunteer' permission (+ managers etc.)
+    form.arrival.choices = generate_arrival_options()
+    form.arrival.data = event_start().strftime("%F")
+    form.departure.choices = generate_departure_options()
+    form.departure.data = event_end().strftime("%F")
 
+    # On sign up give user 'volunteer' permission (+ managers etc.)
     if current_user.is_authenticated and VolunteerUser.get_for_user(current_user):
         return redirect(url_for(".account"))
 
@@ -124,6 +116,8 @@ def account():
         return redirect(url_for(".sign_up"))
 
     form = VolunteerSignUpForm(obj=volunteer)
+    form.arrival.choices = generate_arrival_options()
+    form.departure.choices = generate_departure_options()
 
     if form.validate_on_submit():
         update_volunteer_from_form(volunteer, form)
@@ -136,3 +130,56 @@ def account():
     form.allow_comms.data = volunteer.allow_comms_during_event
 
     return render_template("volunteer/account.html", user=current_user, form=form)
+
+
+def generate_arrival_options():
+    choices = []
+
+    # Work out our first arrival based on config
+    first_arrival = event_start() - timedelta(app.config["ARRIVAL_DAYS"])
+
+    # Add the first date to the list with or earlier
+    choices.append((
+        first_arrival.strftime("%F"),
+        first_arrival.strftime("%A %-d %B or earlier")
+    ))
+
+    # Work out dates between first arrival and end of the event
+    days = period(
+        first_arrival + timedelta(days=1),
+        event_end()).range("days", 1)
+
+    # Add each date to the list
+    for d in days:
+        choices.append((
+            d.strftime("%F"),
+            d.strftime("%A %-d %B")
+        ))
+
+    return choices
+
+def generate_departure_options():
+    choices = []
+
+    # Work out our last arrival based on config
+    last_departure = event_end() + timedelta(app.config["DEPARTURE_DAYS"])
+
+    # Work out dates between start of the event and last departure
+    days = period(
+        event_start(),
+        last_departure - timedelta(days=1)).range("days", 1)
+
+    # Add each date to the list
+    for d in days:
+        choices.append((
+            d.strftime("%F"),
+            d.strftime("%A %-d %B")
+        ))
+
+    # Add last departure day with or later
+    choices.append((
+        last_departure.strftime("%F"),
+        last_departure.strftime("%A %-d %B or later")
+    ))
+
+    return choices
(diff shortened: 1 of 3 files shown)


apps/volunteer/sign_up.py
```diff
@@ -81,6 +79,47 @@ def update_volunteer_from_form(volunteer, form):
 @volunteer.route("/sign-up", methods=["GET", "POST"])
 @feature_flag("VOLUNTEERS_SIGNUP")
 def sign_up():
+
+    # Work out our first and last arrivals based on config
+    first_arrival = event_start() - timedelta(app.config["ARRIVAL_DAYS"])
+    last_departure = event_end() + timedelta(app.config["DEPARTURE_DAYS"])
+
+    # Add the first date to the list with or earlier
+    ARRIVAL_CHOICES.append((
```
SamLR: This should only run once, not every time this endpoint is called (otherwise every time someone signs up the arrival choices list will expand).

I think this is why there's the `generate_day_options` function.

idnorton: Good catch, I completely missed that in testing.

New commit checks the length of the dates lists to see if they're empty before attempting to populate them.

jayaddison: It could be preferable to assign these values at the same time as they're declared in the module; i.e. something along the lines of:

```
# Declare and initialize arrival and departure lists
ARRIVAL_CHOICES = generate_arrival_choices()
DEPARTURE_CHOICES = generate_departure_choices()
```

That'd ensure that the variables are only calculated once, and makes it easier to add automated test coverage, since the tests won't need to invoke `sign_up` before checking the arrival & departure choices.

This could also help other developers to read and understand the module.  The fact that the variables are uppercase implies that they may be [constants according to Python's style guide](https://www.python.org/dev/peps/pep-0008/#constants).  Another developer who is aware of the uppercase variable convention might assume (incorrectly) that `ARRIVAL_CHOICES = []`, for example, implies that the variable would always contain an empty list.

russss: @jayaddison you missed the discussion where we worked out that wasn't possible because it uses config from the app object which isn't available at module scope. 

It has to run within a view function to use the config. At any rate it's definitely not a performance issue.

jayaddison: Ah, sorry - my mistake; ignore my previous comment.  I'll read up on the discussion before adding anything further.

jayaddison: Findings: `wtforms` recommend implementing dynamically-calculated choices by assigning values to the [choices list](https://github.com/emfcamp/Website/blob/2d5cae652076f7aa8c2defac92ba102ba60a1e6c/apps/volunteer/sign_up.py#L38-L41) (more info: [code example](https://github.com/wtforms/wtforms/issues/185#issuecomment-396902599), [docs](https://wtforms.readthedocs.io/en/stable/fields.html#wtforms.fields.SelectField)).

The docs also mention that the `choices` setting is only evaluated once -- so if we're going to use the `ARRIVAL_CHOICES, DEPARTURE_CHOICES` variables, it'll be important to make sure they're assigned before requests reach the endpoint.

For robustness it'd be nice to prove that valid signup dates are accepted and that invalid ones are rejected (perhaps by copying & modifying an [existing endpoint test](https://github.com/emfcamp/Website/blob/2d5cae652076f7aa8c2defac92ba102ba60a1e6c/tests/test_users_app.py#L25-L39) and modifying it for `/volunteer/sign_up`?).

idnorton: Thanks for the pointer to the dynamic choices docs, hopefully the updated version is a better implementation.

idnorton:
Reworked this to use dynamic data passed in to the SelectFields after the initialisation of the form object which removes the globals and the static declarations.

Hopefully this addresses the outstanding issues.

apps/volunteer/sign_up.py
```diff
@@ -82,8 +70,12 @@ def update_volunteer_from_form(volunteer, form):
 @feature_flag("VOLUNTEERS_SIGNUP")
 def sign_up():
     form = VolunteerSignUpForm()
-    # On sign up give user 'volunteer' permission (+ managers etc.)
+    form.arrival.choices = generate_arrival_options()
+    form.arrival.data = event_start().strftime("%F")
+    form.departure.choices = generate_departure_options()
+    form.departure.data = event_end().strftime("%F")
```
jayaddison: I'm a little new to the `data` attribute for these form fields - could you explain what these assignments are for and how they affect the sign-up experience?

apps/volunteer/sign_up.py
```diff
@@ -23,28 +26,13 @@
...
-DEPARTURE_CHOICES = generate_day_options("2018-08-31", "2018-09-03") + [
-    ("2018-09-04", "Tuesday 4 September or later")
-]
-
-
 class VolunteerSignUpForm(Form):
     nickname = StringField("Name", [Required()])
     volunteer_email = StringField("Email", [Email(), Required()])
     over_18 = BooleanField("I'm at least 18 years old")
     volunteer_phone = StringField("Phone", [Required()])
```
jayaddison: @russss Out of interest, do you know if we could/should switch this over to use a [TelField](https://github.com/emfcamp/Website/blob/e37dcb141ab97c93d9d68e8d3887dbee8cb763fc/apps/common/forms.py#L56)?  (not directly related to the changes in this PR)

apps/volunteer/sign_up.py
```diff
@@ -136,3 +130,56 @@ def account():
...
+    days = period(
+        first_arrival + timedelta(days=1),
+        event_end()).range("days", 1)
+
+    # Add each date to the list
+    for d in days:
+        choices.append((
+            d.strftime("%F"),
+            d.strftime("%A %-d %B")
+        ))
```
jayaddison: There may be a few changes we can make to reduce the line-count of the `generate_arrival_options` and `generate_departure_options` methods.

It looks like there's near-duplication of code as a result of handling the '<date> or earlier' and '<date> or later' special case suffixes.

Here's something I've been experimenting with locally which reduces some of the date formatting duplication:

```
    date = event_start() - timedelta(app.config["ARRIVAL_DAYS"])

    choices.append(render_date_option(date, suffix=" or earlier"))  # calls a shared method
    while date < event_end() - timedelta(days=1):
        date += timedelta(days=1)
        choices.append(render_date_option(date))
```

What do you reckon?

apps/volunteer/sign_up.py
idnorton: `data` sets the value of the Select on display so in this case the first and last days of the event will be the defaults presented.

We're using to display the data elsewhere in the same file. The docs are not great for `SelectField`

apps/volunteer/sign_up.py
idnorton: Possibly, though I'm personally not a fan of reducing line length if it also reduces readability.

This is long form and can probably be refactored to reduce the duplication, but equally I'd like to move on to other issues now this is fundamentally working.

I think this is good enough for now, though opinions may vary ;) If the consensus is that it's not good enough then I'll come back to it again.

apps/volunteer/sign_up.py
jayaddison: Ok, makes sense - that display logic makes sense when `sign_up` is called as a `GET`.  It  looks like `sign_up` also handles form content submission via `POST`.

Could you test (either locally or, preferably by adding some test coverage) that the volunteer arrival & departure dates are populated on form submission?

idnorton: Manually tested and it looks good. Other than populating the choices before the form is rendered that shouldn't have affected anything.

apps/volunteer/sign_up.py
idnorton: I'd say that this is out of scope for a bug fix to automatically calculate dates rather than having them manually hard coded.

apps/volunteer/sign_up.py
```diff
@@ -8,13 +8,16 @@
     Markup,
     render_template_string,
 )
+
+from datetime import timedelta
 from flask_login import current_user
 from wtforms import StringField, SelectField, SubmitField, BooleanField
 from wtforms.validators import Required, Email, ValidationError, Optional
 
 from pendulum import parse, period
```
SamLR: as `parse` is no longer being used we can remove it from imports (black should pick this up if you `run_tests` or install the pre-commit hooks)

apps/volunteer/sign_up.py
russss: Yeah, not in scope. I don't know off hand what TelField does so I can't say if it's appropriate.

apps/volunteer/sign_up.py
```diff
@@ -82,8 +70,12 @@ def update_volunteer_from_form(volunteer, form):
 @feature_flag("VOLUNTEERS_SIGNUP")
 def sign_up():
     form = VolunteerSignUpForm()
-    # On sign up give user 'volunteer' permission (+ managers etc.)
+    form.arrival.choices = generate_arrival_options()
+    form.arrival.data = event_start().strftime("%F")
+    form.departure.choices = generate_departure_options()
+    form.departure.data = event_end().strftime("%F")
 
+    # On sign up give user 'volunteer' permission (+ managers etc.)
```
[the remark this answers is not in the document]
idnorton: Yeah, it didn't make much sense to me either. Not sure if I moved it out of context or if it just didn't make sense before I started.

Moved it a few lines down to match up with where the permission is granted.

apps/volunteer/sign_up.py
idnorton: Removed that and fixed my lack of pre-commit. Whoops, my bad :(

apps/volunteer/sign_up.py
```diff
@@ -82,8 +70,12 @@ def update_volunteer_from_form(volunteer, form):
 @feature_flag("VOLUNTEERS_SIGNUP")
 def sign_up():
     form = VolunteerSignUpForm()
-    # On sign up give user 'volunteer' permission (+ managers etc.)
+    form.arrival.choices = generate_arrival_options()
+    form.arrival.data = event_start().strftime("%F")
```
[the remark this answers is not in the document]
idnorton: Tests seemed to be okay but I can see what you mean. Moved these to just before the render assuggested.

apps/volunteer/sign_up.py
idnorton: Re-instated the `generate_day_options` function and made us of it.

I am unconvinced that it adds anything but consensus seems to be that we want it so here it is.

idnorton:
I believe I've addressed the outstanding issues with this now so if people could re-review I'd appreciate it, thanks :)

apps/volunteer/sign_up.py
```diff
@@ -136,3 +132,52 @@ def account():
...
+    # Work out dates between start of the event and last departure
+    choices = generate_day_options(event_start(), last_departure)
+
+    # Replace last array element with the last date and 'or later'
+    choices[len(choices) - 1] = (
+        last_departure.strftime("%F"),
+        last_departure.strftime("%A %-d %B or later"),
+    )
+
+    return choices
```
jayaddison: If and when `generate_day_options` is refactored as suggested, then you could collapse this down to:

```
def generate_departure_options():
    end = event_end() + timedelta(app.config["DEPARTURE_DAYS"])
    day_before_end = end - timedelta(days=1)

    for date in period(event_start(), day_before_end):
        yield generate_day_option(date)
    yield generate_day_option(end, ' or later')
```

PR #875, opened 2020-02-18
Display EUR bank details as an EPC QR code

### LINKED ISSUE #841: Display EUR bank details as an EPC QR code
https://en.wikipedia.org/wiki/EPC_QR_code

jayaddison:
This change updates the invoice rendering code to include an [EPC QR code](https://www.europeanpaymentscouncil.eu/sites/default/files/KB/files/EPC069-12%20v2.1%20Quick%20Response%20Code%20-%20Guidelines%20to%20Enable%20the%20Data%20Capture%20for%20the%20Initiation%20of%20a%20SCT.pdf) as [rendered by segno](https://segno.readthedocs.io/en/stable/epc-qrcodes.html) for the purposes of fast and accurate payment submission.

These codes are rendered into invoices only when payment is pending as a bank transfer in Euros (EUR).

image

Todo list:

- [x] Remove code duplication between QR & EPC QR generation
- [ ] Test with Transferwise account details
- [ ] Reduce left-padding around QR code?
- [x] Add test coverage

Resolves #841 

[epc_structured_reference.pdf](https://github.com/emfcamp/Website/files/4231247/epc_structured_reference.pdf)

### DIFF
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -110,7 +110,7 @@ def make_epc_qrfile(payment, **kwargs):
         name="Electromagnetic Field Ltd",
         iban="GB21BARC20716472954433",
         amount=payment.amount,
-        text=payment.bankref,
+        reference=payment.bankref,
     )
     qr.save(qrfile, **kwargs)
     qrfile.seek(0)
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -111,6 +111,7 @@ def make_epc_qrfile(payment, **kwargs):
         iban="GB21BARC20716472954433",
         amount=payment.amount,
         reference=payment.bankref,
+        encoding=1,
     )
     qr.save(qrfile, **kwargs)
     qrfile.seek(0)
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -117,9 +117,7 @@ def make_epc_qrfile(payment, **kwargs):
     return qrfile
 
 
-def format_inline_qr(data):
-    qrfile = make_qrfile(data, kind="svg", svgclass=None)
-
+def qrfile_to_svg(qrfile):
     root = etree.XML(qrfile.read())
     # Allow us to scale it with CSS
     root.attrib["viewBox"] = "0 0 %s %s" % (root.attrib["width"], root.attrib["height"])
@@ -128,19 +126,16 @@ def format_inline_qr(data):
     root.attrib["preserveAspectRatio"] = "none"
 
     return Markup(etree.tostring(root).decode("utf-8"))
+
+
+def format_inline_qr(data):
+    qrfile = make_qrfile(data, kind="svg", svgclass=None)
+    return qrfile_to_svg(qrfile)
 
 
 def format_inline_epc_qr(payment):
     qrfile = make_epc_qrfile(payment, kind="svg", svgclass=None)
-
-    root = etree.XML(qrfile.read())
-    # Allow us to scale it with CSS
-    root.attrib["viewBox"] = "0 0 %s %s" % (root.attrib["width"], root.attrib["height"])
-    del root.attrib["width"]
-    del root.attrib["height"]
-    root.attrib["preserveAspectRatio"] = "none"
-
-    return Markup(etree.tostring(root).decode("utf-8"))
+    return qrfile_to_svg(qrfile)
 
 
 def make_qr_png(url):
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -7,6 +7,7 @@ from pyppeteer.launcher import launch
 import barcode
 from barcode.writer import ImageWriter, SVGWriter
 import segno
+from segno import helpers
 
 from main import external_url
 from models import event_year
@@ -103,8 +104,34 @@ def make_qrfile(data, **kwargs):
     return qrfile
 
 
+def make_epc_qrfile(payment, **kwargs):
+    qrfile = io.BytesIO()
+    qr = helpers.make_epc_qr(
+        name="Electromagnetic Field Ltd",
+        iban="GB21BARC20716472954433",
+        amount=payment.amount,
+        reference=payment.bankref,
+    )
+    qr.save(qrfile, **kwargs)
+    qrfile.seek(0)
+    return qrfile
+
+
 def format_inline_qr(data):
     qrfile = make_qrfile(data, kind="svg", svgclass=None)
+
+    root = etree.XML(qrfile.read())
+    # Allow us to scale it with CSS
+    root.attrib["viewBox"] = "0 0 %s %s" % (root.attrib["width"], root.attrib["height"])
+    del root.attrib["width"]
+    del root.attrib["height"]
+    root.attrib["preserveAspectRatio"] = "none"
+
+    return Markup(etree.tostring(root).decode("utf-8"))
+
+
+def format_inline_epc_qr(payment):
+    qrfile = make_epc_qrfile(payment, kind="svg", svgclass=None)
 
     root = etree.XML(qrfile.read())
     # Allow us to scale it with CSS
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -119,8 +119,7 @@ def make_epc_qrfile(payment, **kwargs):
 
 
 def qrfile_to_svg(qrfile):
-    root = etree.XML(qrfile.read())
-    return Markup(etree.tostring(root).decode("utf-8"))
+    return Markup(qrfile.getvalue().decode("utf-8"))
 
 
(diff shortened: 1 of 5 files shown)


jayaddison:
Testing unstructured EPC remittance information field:

image

davidc:
Structured (35 chars field) -
Deutsche Bank: Inputs to Description (140 characters), customer reference (35 characters) is blank. (Seems like DB might be using the wrong field here)
Sparkasse: Inputs to Verwendungszweck

Unstructured (140 chars field) -
Deutsche Bank: does not appear at all
Sparkasse: does not appear at all

kliment:
Dutch ING app will not scan first code. With the second code reference shows up in description field

jayaddison:
Thanks @Walther!

> MobilePay (a somewhat commonly used 3rd party service): might be expecting only its own special kind of QR codes.
> 
> IMG_49FFA9E71DCB-1

cc @heuer - it looks like the [MobilePay](https://en.wikipedia.org/wiki/MobilePay) service didn't recognize this image generated via `segno` for some reason, and reported it as invalid.

The emfcamp website's code takes the output of [segno.helpers.make_epc_qr](https://segno.readthedocs.io/en/stable/api.html#segno.helpers.make_epc_qr) and converts the resulting image into an SVG, so the root cause may lie either inside or outside the segno library.  I'll try to debug a little further when time allows.

squelcher:
Triodos Bank Germany, using MyBankingApp (Fiducia & GAD IT AG)

Both: Recipient and IBAN show up correctly, no BIC/SWIFT code appears.

Structured Reference: VFH9K3RQ (without -) shows up as "Kundenreferenz" ("Customer Reference")

Unstructured Reference: VFH9K3RQ (without -) shows up as "Verwendungszweck" (reason for transfer)

In Germany, we rarely use the "Kundenreferenz" field, the much longer "Verwendungszweck" is mainly used for that.

jayaddison:
Thanks everyone for the help with testing this!  Based on the varying result discovered here, this functionality requires a bit more care and attention before it's ready, and I'll resume work on it over the next few days.

heuer:
Thanks for using Segno. ;)

Maybe some scanners like MobilePay don't like the ISO-8859-1 encoding and expect UTF-8?
By default, ``make_epc_qr`` uses UTF-8 if all other encodings fail. Maybe you can try to enforce UTF-8 encoding by setting the encoding parameter to 1.

```
qr = helpers.make_epc_qr(..., encoding=1)
```
_(By the way, it would have been much better to specify the encoding with encoding='utf-8')_ 

If I can do anything in this matter, let me know. In case you've found a bug, add an issue to the tracker.

apps/common/receipt.py
```diff
@@ -130,17 +128,14 @@ def format_inline_qr(data):
     return Markup(etree.tostring(root).decode("utf-8"))
 
 
-def format_inline_epc_qr(payment):
-    qrfile = make_epc_qrfile(payment, kind="svg", svgclass=None)
+def format_inline_qr(data):
+    qrfile = make_qrfile(data, kind="svg", svgclass=None)
```
heuer: Maybe you should use:
```
qrfile = make_qrfile(data, kind='svg', svgclass=None, omitsize=True)
```
Since you're replacing the height / width attributes a viewBox in line 125 anyway.
See also <https://segno.readthedocs.io/en/stable/svg-options.html>

heuer:
BTW, I used BCD Checker <https://www.stuzza.at/en/download/qr-code.html> (only available for Windows but I ran it with Wine) to check my implementation. 

Acc. to BCD Checker both EPC QR codes are fine.

apps/common/receipt.py
jayaddison: Brilliant, that allows for a nice bit of code cleanup.  Thanks!

jayaddison:
> Thanks for using Segno. ;)

Thanks for developing the library :)
 
> Maybe some scanners like MobilePay don't like the ISO-8859-1 encoding and expect UTF-8?

That sounds worth investigating, yep!  Thanks for the pointer.  I've pushed a code change for this.

@kliment Does the Dutch ING app now scan the (updated, UTF-8 version) first image?

@Walther Similarly, does MobilePay seem any happier with the new updated images here?

kliment:
My Dutch friend says ING still won't accept the first code.

apps/common/receipt.py
```diff
@@ -103,19 +104,35 @@ def make_qrfile(data, **kwargs):
...
 
-    root = etree.XML(qrfile.read())
-    # Allow us to scale it with CSS
-    root.attrib["viewBox"] = "0 0 %s %s" % (root.attrib["width"], root.attrib["height"])
-    del root.attrib["width"]
-    del root.attrib["height"]
-    root.attrib["preserveAspectRatio"] = "none"
 
+def qrfile_to_svg(qrfile):
+    root = etree.XML(qrfile.read())
```
heuer: I guess the XML parsing is unecessary, this should work as well:

```python
return Markup(qrfile.getvalue().decode('utf-8'))
```
If you make the detour via lxml to get rid of the XML declaration, you can also tell Segno to omit it (line 127 and 132):

```python
# Omit XML declaration and trailing new line 
qrfile = make_qrfile(data, kind="svg", svgclass=None, omitsize=True, xmldecl=False, nl=False)
```

If the QR code is used in a HTML5 context, the SVG namespace declaration may be superfluous as well:
```python
# Omit XML declaration, SVG namespace and trailing new line 
qrfile = make_qrfile(data, kind="svg", svgclass=None, omitsize=True, xmldecl=False, svgns=False, nl=False)
```

jayaddison:
@kliment Okie doke, no worries.  I'll start to look into #878 and related issues as a next potential fix before we try again.

Walther:
Nordea still scans both, MobilePay still complains about invalid QR

apps/common/receipt.py
jayaddison: Neat; taking a cautious approach here before dropping the XML and namespace declarations, but removing the XML parsing certainly seems a safe & nice improvement to start with.

heuer: No problem, but I guess dropping the XML declaration is safe because
```
etree.tostring(root).decode("utf-8")
```
returned the XML markup without an XML declaration anyway.

heuer:
>     * Reduce left-padding around QR code?

If you use border=0 in the QRCode.save() method, the border will be omitted (on all sides). Please note that the ISO standard recommends a border (quiet zone) of 4 modules. If there is enough white space around the QR Code I don't see any problem to omit the border, though.

jayaddison:
This is confusing, but I forgot that this PR was open before deleting the forked repo this branch was created from.  That means I couldn't push any further commits via GitHub; I'm going to reopen this momentarily.

jayaddison:
@davidc @Walther @squelcher @kliment thank you again for testing this a few years ago.  If / when you have time: could you check whether the updated image below scans better?

image

(:warning: the destination IBAN is a test value -- but even so, please don't attempt to send any funds)

The above was generated with code from #1292, including a fix for the format of structured references.

jayaddison:
...and here's one more update, with the amount configured down to 1EUR for safety (something I should have done previously), and with a SWIFT code -- still a test value, but updated to correspond to the IBAN -- included in the EPC QR:

image

Again, despite the reduced amount and test IBAN, please don't initiate actual payments to the destination.

Walther:
<details>
<summary>iPhone camera</summary>

IMG_3115

</details>

<details>
<summary>Nordea Mobile</summary>

IMG_3116

</details>

MobilePay refuses to acknowledge the QR code. There's no error anymore, it just doesn't appear to scan at all - probably made to ignore unknown codes instead of throwing in some update.

And apologies, I should have checked this earlier: my earlier hunch was correct. According to their own documentation, MobilePay only supports QR codes for points-of-sale use when using their own format: https://github.com/MobilePayDev/MobilePayDev.github.io/blob/main/docs/pos/create-qr-codes.md

Sorry about the extra work! 🙇‍♂️

davidc:
Seems fine, this is what shows up in the Deutsche Bank app. I assume the bank name being different is intended.

IMG_9086

IMG_9087

IMG_9088

jayaddison:
@Walther @davidc thank you both!

@Walther excellent, Nordea looks much better.  And that's no problem about MobilePay - that's good to be aware of.

@davidc
> I assume the bank name being different is intended.

Thanks again - that is expected, but I could and should update the example data so that the test SWIFT/BIC code and bank name are consistent with each other.

I've since learned that when the [8th character of a BIC code is zero](https://en.wikipedia.org/wiki/ISO_9362#Structure), it is generally considered a "test and training" code -- I'm considering switching the example BIC to use that instead, although I'm unsure how banking apps will handle that.

PR #876, opened 2020-02-18
Dynamically display the next sales date

jellybob:
I'm in no way proud of the hack to get the timezone into the date, but it works for this very specific purpose, and I was disappearing into a rabbit hole of Python timezone documentation.

Requires `SALES_DATES` to be added to the config as shown in `development-example.cfg`

### DIFF
--- a/apps/base/__init__.py
+++ b/apps/base/__init__.py
@@ -17,6 +17,7 @@ from flask import (
 
 from main import cache
 from ..common import feature_flag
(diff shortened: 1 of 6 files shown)


russss:
I have to say I'm not a huge fan of this - I think we can probably handle it manually. The system is already quite complex and I think if we're adding more state like this it will need to be in the database and configurable from the admin interface.

PR #881, opened 2020-02-20
Minor HTML templating fixups

jayaddison:
These don't appear to have affected rendering but seemed worth cleaning up while editing the invoice template.


jayaddison:
(may be easiest to review with [whitespace changes collapsed](https://github.com/emfcamp/Website/pull/881/files?w=1))

PR #884, opened 2020-02-22
Add model and view for management of training

idnorton:
The admin UI is sub optimal for the longer term however it delivers enough to be able to add training, questions and answers ready for working on the user facing aspects which are the focus now.


models/volunteer/training.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @classmethod
+    def get_by_id(cls, id):
+        return cls.query.get(id)
+
+    @classmethod
+    def get_by_name(cls, name):
+        return cls.query.filter_by(name=name).one_or_none()
+
+class TrainingQuestion(db.Model):
```
SamLR: do the training questions/answers need to be modelled in the database? I feel they could probably just be a JSON blob in the training

russss:
Just so this is in the same place, @SamLR said:

`11:38 PM <SamLR> idn: for the training questions/answers is there a reason to model them in the database at all? Feels like something that we could probably just store as a blob of json`

I think I agree with this. From my perspective I think the only situation where we're going to have this kind of online training quiz is for the bar staff (because it is a legal requirement). I think it would be excessive to do this for anything else.

Given that we wipe the database after each event, I don't really see any reason why this stuff needs to be in the DB as opposed to a static JSON file.

idnorton:
The reason for this was a comment from @Jonty :

> 16:54 < jontyw> Brief note that we _may_ want to have similar training capabilities for other roles, so something vaguely generic might be useful.

So I took the generic approach from the start. I guess there's no requirement to have them in the DB but it felt like the right approach. Do we want to switch approach to a JSON blob in a file instead at this point or does this requirement change your thoughts?

sde1000:
> I think I agree with this. From my perspective I think the only situation where we're going to have this kind of online training quiz is for the bar staff (because it is a legal requirement). I think it would be excessive to do this for anything else.

And for the bar training, we don't need to store people's answers: we just need to store that they have completed the training (and ensure that to complete the training you must answer the questions correctly).

idnorton:
We're not going with this approach

Issue #857, opened 2020-01-22
Switch bank transfer payments to use Transferwise


russss:
Looks like the [Transferwise API](https://api-docs.transferwise.com/) is available for us to use now. It would be good to use Transferwise for all incoming bank payments, as they have a proper API.

We can automatically create the bank accounts in our DB by calling [get account balance](https://api-docs.transferwise.com/#borderless-accounts-get-account-balance) - probably needs to be an admin job to do that.

Then we need to listen for the [balance deposit event](https://api-docs.transferwise.com/#webhook-events-balance-deposit-event) webhook, fire off a [get account statement](https://api-docs.transferwise.com/#borderless-accounts-get-account-statement) request for that bank account, import any un-imported transactions, then reconcile.

This will be (possibly much) faster, remove the dependency on my bank scraping infrastructure, and fix #570.

jayaddison:
Here's my current understanding of the issue - much of this will just be echoing the description to see whether I've understood it correctly.

The plan is to listen for incoming balance credits to a receiving TransferWise bank account, by using a webhook subscription created for the TransferWise profile.  This subscription will listen for `balances#credit` events (the only event type we need to listen for currently).

When a credit is received, we'll request[q1] a recent account statement using the `profileId` from the webhook and `borderlessAccountId` for the account[q2].

For each transaction in the response, we'll find a corresponding `BankPayment` object (if any) based on the `paymentReference`, validate the payment amount and details, and transition the payment into an appropriate state.

Questions:

1. What time interval should the account statements be requested for - and will this be based on a 'high water mark' timestamp of previous successfully processed transactions, or should we just request the last `n` days each time?  (max `n` == 469 days according to the TransferWise docs)
1. Where will the `borderlessAccountId` be modelled in the DB?
1. More generally, how are the relevant bank accounts modelled in the DB, and/or what will be required to enable an admin task for this? (cc @russss)

russss:
> For each transaction in the response, we'll find a corresponding BankPayment object (if any) based on the paymentReference, validate the payment amount and details, and transition the payment into an appropriate state.

There are actually two steps here:

1. Assuming the payment is a credit (we ignore debits entirely), first we insert the transaction as a [BankTransaction](https://github.com/emfcamp/Website/blob/877510cab3626fb83a71bb7a24cb574d891b83a7/models/payment.py#L325), if it doesn't already exist.
2. Then we try and reconcile the transaction with existing BankPayments.

[...]

PR #888, opened 2020-03-04
Updates the BankAccount model and related functionality

### LINKED ISSUE #570: Store bank details in BankAccount and obfuscate fixture
To get the DD details out of cleartext in git. We can probably just base64 the .json

jayaddison:
This changeset continues to work towards support for TransferWise as discussed in #857.

The following changes are applied:

- SWIFT and IBAN required fields are added to the `BankAccount` model
- The existing sort code and account number fields on `BankAccount` are retained, but become optional
- An optional `active` flag is added to the `BankAccount` model -- along with a corresponding `UNIQUE` constraint to ensure that only a single account may be marked as active per-currency
- ~~The CLI command to load bank accounts is removed~~ (rollback: currently still required for unit tests)
- The payments verification admin page is extended to list the bank accounts currently available
- User-facing templates are updated to render bank account details from `BankAccount` objects instead of static text

Resolves #570


migrations/versions/d01b87ff85d2_internationalize_bank_account_fields.py
```diff
@@ -0,0 +1,64 @@
...
+               nullable=True)
+    op.alter_column('bank_account', 'sort_code',
+               existing_type=sa.VARCHAR(),
+               nullable=True)
+    op.create_index('ix_bank_account_currency_active', 'bank_account', ['currency', 'active'], unique=True)
+    # ### end Alembic commands ###
+
+    # manual custom operations #
+    op.execute("UPDATE bank_account SET active = true")
+    op.execute("UPDATE bank_account SET iban = CASE WHEN acct_id LIKE '%0526' THEN 'GB39BARC' ELSE 'GB21BARC' END || acct_id")
```
jayaddison: :warning: Although these custom migration script steps have been tested for backwards-and-forwards compatibility in development (i.e. repeated `flask db upgrade` ... `flask db downgrade` both work and leave the DB and data in the expected state), this `UPDATE` operation on L35 assumes that only the two 'current' bank accounts exist when it runs.

Technically the script could perform an ISO-7064 checksum to figure out the IBAN check digs (the 'XX' in 'GBXXBARC'), but given only two existing accounts... it's easier simply to use a conditional check.

templates/emails/tickets-reminder.txt
```diff
@@ -24,21 +24,21 @@ ensure it's included in full.
...
 Amount:         {{ payment.amount | price('GBP') }}
 Reference:      {{ payment.bankref | bankref }}
 
 (Don't worry if the payee name doesn't quite fit)
 
-{% elif payment.currency == 'EUR' %}
+{% else %}
 
-You can pay into our Euro bank account:
+You can pay into our {{ account.currency }} bank account:
```
jayaddison: A small note highlighting the changes around `currency != GBP` in the diff - these are worth a bit of scrutiny during review.

models/payment.py
```diff
@@ -295,19 +295,39 @@ def manual_refund(self):
...
+    sort_code = db.Column(db.String)
+    acct_id = db.Column(db.String)
     currency = db.Column(db.String, nullable=False)
-
-    def __init__(self, sort_code, acct_id, currency="GBP"):
+    active = db.Column(db.Boolean)
+    institution = db.Column(db.String, nullable=False)
+    address = db.Column(db.String, nullable=False)
+    swift = db.Column(db.String, nullable=False)
+    iban = db.Column(db.String, nullable=False)
```
russss: I think it would probably be best to make SWIFT and IBAN nullable - although our existing Barclays GBP account does technically have an IBAN, it's not something we'd ever expect to use (it causes expense if someone attempts to send euros to it), so it's probably better if we don't worry about it.

I think this would also slightly simplify the migration.

jayaddison:
@russss I'm closing and/or merging some old pull requests before the holidays and I think this one is reasonable to merge.  It's a bit stale but I think we were mostly comfortable with it at the time; I'll get conflicts and tests resolved and merge unless there are any concerns.

jayaddison:
Merging; the GitHub review on this is technically still open/pending but has been addressed.

russss:
Thanks for your help Jay!

Hopefully I'll have some time to review things soon, and reprioritise what needs to happen next.

PR #889, opened 2020-03-04
Admin page: Associate 'Manual tasks' block with GoCardless

jayaddison:
Relocate the manual tasks which relate to GoCardless to sit underneath the relevant heading on the `payment-config-verify` admin page.  This is to reduce a little bit of visual confusion and felt worthwhile during some page edits in other branches.


russss:
Thanks, I spotted that and was going to fix it myself at some point.

PR #892, opened 2020-03-04
Tests: don't create fixtures or run the web app

marksteward:
To stop the tests silently failing after an undefined number of seconds if the web app or fixtures themselves throw errors.


jayaddison:
Tested via a known-bad [commit](https://github.com/emfcamp/Website/pull/888/commits/4086e3c37bfa42dec75f5e93418c5eb1d78389fd) - tests now run to completion as expected and the failure is visible in the output: https://github.com/emfcamp/Website/runs/487584143

PR #893, opened 2020-03-10
Handle incoming Wise webhooks

jayaddison:
Adds a payments endpoint at `/transferwise-webhook` which receives POSTed Wise [webhook events](https://api-docs.transferwise.com/#webhook-events).

On receipt of balance credits, we retrieve recent transaction statements for the corresponding Wise account, and attempt to match each transaction against a `BankTransaction` in the database (or construct one if none exist).

For each resulting `BankTransaction`, a reconciliation process finds corresponding `Payment` entries and, where valid, marks them as paid and issues customer notifications.

Todo:
- [ ] Reconcile payments (extract [common functionality](https://github.com/emfcamp/Website/blob/7e839497fcdc3cc27c27c936009e07008cc41004/apps/base/tasks_banking.py#L139-L215) from `reconcile` CLI command?)
- [ ] Test coverage

Relates to #857

### DIFF
--- a/apps/payments/transferwise.py
+++ b/apps/payments/transferwise.py
@@ -1,7 +1,53 @@
-from flask import current_app as app
+from flask import abort, current_app as app, request
 import pywisetransfer
+from pywisetransfer.webhooks import verify_signature
+import logging
 
+from main import csrf
 from models.payment import BankAccount
+from . import payments
+
+logger = logging.getLogger(__name__)
+
+
+webhook_handlers = {}
+
+
+def webhook(type=None):
+    def inner(f):
+        webhook_handlers[type] = f
+        return f
+
+    return inner
+
+
+@csrf.exempt
+@payments.route("/transferwise-webhook", methods=["POST"])
+def transferwise_webhook():
+    valid_signature = verify_signature(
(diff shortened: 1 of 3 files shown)


apps/payments/transferwise.py
```diff
@@ -1,7 +1,121 @@
...
+    txns = []
+    for transaction in statement.transactions:
+        if transaction.type != "CREDIT":
+            continue
+
+        # Attempt to find transaction in the application database
+        txn = BankTransaction.query.filter(
+            account_id=bank_account.id,
+            posted=transaction.date,
+            type=transaction.details.type.lower(),
```
jayaddison: From the TransferWise [borderless account statement](https://api-docs.transferwise.com/#borderless-accounts-get-account-statement) docs:

```
transactions[n].details.type
---
CARD, CONVERSION, DEPOSIT, TRANSFER, MONEY_ADDED
```

This will have an impact on [transaction type validation](https://github.com/emfcamp/Website/blob/7e839497fcdc3cc27c27c936009e07008cc41004/apps/base/tasks_banking.py#L140) during reconciliation (assuming code is re-used).

Issue #867, opened 2020-02-16
Volunteer training UI / bar training


idnorton:
Primarily to handle the Bar training, the model and basic view for dealing with volunteer training has been added but we now need UI to handle a volunteer going through the process of completing training.

We have the following in the model:

* Training
* TrainingQuestion
* TrainingAnswer

For a given set of training, it can have one or more questions which can have one or more valid answer.

Depending on the requirements of the training, the test can automatically sign someone off based on a pass mark, or require someone to manually confirm training completed.

idnorton:
So we discussed this issue on PR #884 and the consensus of opinion was that there was no requirement to have online editing of the training, questions and answers and it would be preferable to have those stored in a static JSON file in source control than data driven in the db.

Here's a first stab at a json data structure based on the 2018 training questions and answers.
[training.txt](https://github.com/emfcamp/Website/files/4247126/training.txt)

Comments and thoughts on this? It's basically a lift and shift of the data I manually entered into my local database instance and feels about right.

idnorton:
Comments and feedback discussion happened on IRC and I forgot to add them here:
```
--- Day changed Wed Feb 26 2020
11:46 < idn> Any comments or thoughts on the suggested JSON format in https://github.com/emfcamp/Website/issues/867 peeps?
11:48 <@ms7821> I like it
11:51 < idn> Was contemplating one file per training thing, which is only bar at the moment but could be more later.
11:51 < idn> Then keeps things self contained.
11:52 <@ms7821> yeah that would be good
11:52 < idn> Also tagging the correct answer means that we could potentially have one or more valid answers with checkboxes rather than just a single choice drop down
11:52 <@ms7821> yep
11:53 < idn> Cool beans, I'll try and do something with that in the next week then
11:53 < idn> Thanks for the sanity check ms7821
11:53 < idn> (this is basically a lift and shift of the db model I did)
12:38 < jontyw> I like it too
12:39 < jontyw> And this comes from someone who has professionally studied basically all available survey tool export/import formats
12:39 < jontyw> :P
12:47 < idn> jontyw: Thanks :)
12:47 < ormiret> idn: one comment, the correct (and pass_auto) flags are strings in that, "false" is truthy
12:48 < idn> ormiret: Yeah, my JSON isn't great so that's probably poor understanding on my part. What should it be/
12:49 < ormiret> without the quotes would get bools
12:49 < jontyw> Yup
[...]

PR #920, opened 2021-02-03
Adding fixture for bar training

idnorton:
Further to discussions in #867 here's a first stab at adding the fixture model for bar training.

### DIFF
--- a/models/fixtures/training/bar.json
+++ b/models/fixtures/training/bar.json
@@ -7,6 +7,125 @@
       "url": "",
       "questions": [
         {
+          "question": "At what percentage of abv is a drink legally classed as alcohol?",
+          "answers": [
+            {
+              "answer": "More than 0.1%",
+              "correct": false
+            },
+            {
+              "answer": "More than 0.3%",
+              "correct": false
+            },
+            {
+              "answer": "More than 0.5%",
+              "correct": true
+            }
+          ]
+        },
+        {
+          "question": "How long does it take the body to remove one unit of alcohol?",
+          "answers": [
+            {
+              "answer": "30 minutes",
+              "correct": false
+            },
+            {
+              "answer": "60 minutes",
+              "correct": true
+            },
+            {
+              "answer": "90 minutes",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "How many units of alcohol are in a 50ml measure of spirit that has an ABV of 40%?",
+          "answers": [
+            {
+              "answer": "1",
+              "correct": false
+            },
+            {
+              "answer": "2",
+              "correct": true
+            },
+            {
+              "answer": "3",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "Which of these is <strong>not</strong> a licensing objective?",
+          "answers": [
+            {
+              "answer": "Prevention of crime and disorder",
+              "correct": false
+            },
+            {
+              "answer": "Protection of the environment",
+              "correct": true
+            },
+            {
+              "answer": "Public safety",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "On licensed premises, a designated premises supervisor is someone who is:",
+          "answers": [
+            {
+              "answer": "Always required to be present on particular licensed premises",
+              "correct": false
+            },
+            {
+              "answer": "In day-to-day control of a particular licensed premises",
+              "correct": true
+            },
+            {
+              "answer": "Required to personally supervise every sale of alcohol on licensed premises",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "What must an operating schedule include?",
+          "answers": [
+            {
+              "answer": "Addresses and telephone numbers of all staff at the premises",
+              "correct": false
+            },
+            {
+              "answer": "Details of the licensable activities to be held at the premises",
+              "correct": true
+            },
+            {
+              "answer": "Names of all suppliers of alcohol and food to the premises",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "When is it <strong>illegal</strong> to sell alcohol?",
+          "answers": [
+            {
+              "answer": "Outside the times stated in the premises licence",
+              "correct": true
+            },
+            {
+              "answer": "Under the terms of a temporary event notice",
+              "correct": false
+            },
+            {
+              "answer": "When the designated premises supervisor is not present",
+              "correct": false
+            }
+          ]
+        },
+        {
           "question": "A proof of age document is reliable if it has",
           "answers": [
             {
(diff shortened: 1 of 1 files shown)


models/fixtures/training/bar.json
```diff
@@ -0,0 +1,93 @@
...
+              "correct": true
+            },
+            {
+              "answer": "Police authority stamp",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "What is the on-the-spot fine for selling alcohol to someone who is under 18?",
```
jayaddison, line 27: To help review (current and future) and maintenance, can we include some supporting references for each question?  (perhaps via a hyperlink or two in a commit message)

models/fixtures/training/bar.json
```diff
@@ -0,0 +1,93 @@
...
+              "correct": false
+            },
+            {
+              "answer": "25",
+              "correct": false
+            }
+          ]
+        },
+        {
+          "question": "Which of these is considered a common sign of drunkenness?",
```
jayaddison, line 74: It's good that we require 100% to pass the test, but I worry that this question might be a little under-leveled.  It's useful as a pre-screen but I don't know if it'd stand up to strong scrutiny of the training.

models/fixtures/training/bar.json
```diff
@@ -0,0 +1,93 @@
+{
+  "training": [
+    {
+      "role": "Bar",
+      "pass_auto": true,
+      "pass_mark": 100,
+      "url": "",
```
jayaddison, line 7: Is there a missing value here?

jayaddison (changes requested):

models/fixtures/training/bar.json
sde1000: There are more questions in the full version of the test. I assume this file is a cut-down version for, er, testing.

models/fixtures/training/bar.json
idnorton: Yes, I'm not sure we know what's happening with the actual training material, that has previously been hosted elsewhere.

models/fixtures/training/bar.json
idnorton: This isn't the training material, this is purely the test to confirm someone has read and understood the training material.

I might be having a lack of memory on the details at this point but I'm not sure if there were more than 3 questions last time?

idnorton: Ah, no, there's 5 here. I'n pretty sure that's what we had last time.

models/fixtures/training/bar.json
```diff
@@ -0,0 +1,93 @@
+{
+  "training": [
+    {
+      "role": "Bar",
+      "pass_auto": true,
```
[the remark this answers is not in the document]
idnorton: Sure, as soon as there's something to document.

models/fixtures/training/bar.json
idnorton: When originally discussing this the URL was a config setting and I was told it was going to be rewritten:
```
16:13 < idn> Bar training is exploding at the moment, I have no idea what to do with that, but that's just me randomly clicking around. Should I card that?
16:13 < idn> KeyError: 'BAR_TRAINING_FORM'
16:14 < idn> Ah, missing config perhaps.
16:14 < russss> might be config again yeah...
16:14 < idn> Yeah, looks like it - url = app.config["BAR_TRAINING_FORM"]...
16:15 <@ms7821> bar training will be rewritten
16:15 < idn> ms7821: k, I'll ignore that then.
16:15 < idn> Ta :)
```
So at this point we don't have a URL to add.

models/fixtures/training/bar.json
idnorton: I stand corrected, there were 12 answers in the original source:

https://github.com/emfcamp/Website/blob/master/apps/volunteer/bar_training.py#L79

So yes this is a partial list of both questions and answers. I'll add a comment to the ticket to make that clear.

models/fixtures/training/bar.json
sde1000: I would like the training material to be hosted on this website for next time. See "Training outline" in the Bar folder of the 2018 google drive. It may change before next time, it depends how the licence application process goes.

idnorton: Yup, but right now I'm focused on the test part.

idnorton: @sde1000 I don't think I have access to the Google Drive folder of which you speak.

idnorton: I'm assuming it's the same as this one though - https://usercontent.irccloud-cdn.com/file/3LYiO0Er/Training%20outline.pdf

sde1000: It is.

models/fixtures/training/bar.json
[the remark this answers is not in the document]
jayaddison: No problem - I think the IRC discussion from https://github.com/emfcamp/Website/issues/867#issuecomment-772840906 covers it.

models/fixtures/training/bar.json
jayaddison: (update: I don't think we need a reference for each individual question, if they're all sourced from the same place.  the [original bar training document](https://github.com/emfcamp/Website/pull/920#discussion_r575707621) would be fine)

models/fixtures/training/bar.json
jayaddison: Thanks @sde1000 @idnorton!

models/fixtures/training/bar.json
jayaddison: (...I think.  just checking what the retention policy for IRCCloud CDN documents is...)

idnorton: @jayaddison This is short term so next steps after this is likely migration of that content to something within the website, at which point the referencing from URL will hopefully make more sense.

idnorton: Starting to think that perhaps I should rename the file to be bar_test.json and then we can have bar_content.json or similar. That sound like a sensible plan?

sde1000: I'll be reviewing and updating the training ahead of the 2022 event. I would like the training material to be hosted on this website; when it is, this repo can become the original document.

jayaddison: (uncertain so far; but durable enough for the emfcamp website, I reckon)

models/fixtures/training/bar.json
jayaddison: If we could place a link to the original bar training document into this `url` JSON field (even if we change/remove it later), that'd keep me happy that we have a git-based record/reference to the source material.

jayaddison (approved):
lgtm!

PR #930, opened 2021-06-30
Add __str__ and __repr__ for Volunteer Model

ormiret:


models/volunteer/volunteer.py
```diff
@@ -63,6 +63,12 @@ class Volunteer(db.Model, UserMixin):
         lazy="dynamic",
     )
 
+    def __repr__(self):
+        return "<Volunteer {0}<{1>>".format(self.user.name, self.user.email)
```
russss, line 67: You've got a > instead of a } there.

(I'd probably use [f-strings](https://docs.python.org/3/tutorial/inputoutput.html) for this now - they're supported on our python version and I think they'd be a bit clearer)

russss (changes requested):

models/volunteer/volunteer.py
ormiret: Updated to f-strings now

idnorton:
Looks good to me, tests are now passing.

russss (approved):

PR #936, opened 2021-08-10
Make it so under 18s can sign up for volunteer roles (fixes #934)

### LINKED ISSUE #934: Under 18s don't see "Sign up for this role" button in more details page of a role
In #933 we discovered that if the volunteer is under 18, you don't see the button in the UI.

Steps to reproduce:

1. Create volunteer which is under 18 (or change your existing)
2. Go to roles page https://www.emfcamp-test.org/volunteer/choose-roles and click the "More" shown at the end of description
3. Should see the following but do not:

image

We should see this button if the role is able to be performed by someone under 18.

anorthall:
As above, and per issue #934, this should ensure that over 18s can sign up for all roles, and under 18s can only sign up for roles that are applicable to them.

### DIFF
--- a/templates/volunteer/role.html
+++ b/templates/volunteer/role.html
@@ -27,7 +27,17 @@
         {{ description }}
     {% endif %}
 
-    {% if current_volunteer.over_18 %}
+    {% if role.over_18_only %}
+        {% if current_volunteer.over_18 %}
+            <form class="form-inline" method="POST">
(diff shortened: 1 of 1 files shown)


idnorton:
This is a good fix for ensuring that over 18s are only presented the sign up button for roles that require it, but it doesn't give us the button for under 18s when that's not required? Unless I'm missing something here?

PR #937, opened 2021-08-10
Update mailing list sign up to read 2022 instead of 2020

anorthall:
I don't think this needs any further explanation!

### DIFF
--- a/templates/home/before-sales.html
+++ b/templates/home/before-sales.html
@@ -27,7 +27,7 @@
          January 2018 at 19:00</strong> (UK time).</p>
       #}
        <div class="well">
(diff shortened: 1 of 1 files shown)


idnorton (approved):
:+1:

idnorton (changes requested):
We had a quick chat about this on IRC and would like to get this moved to a template date something akin to `{{event_start.year}}` if that's possible?

Thanks :)

anorthall:
> We had a quick chat about this on IRC and would like to get this moved to a template date something akin to `{{event_start.year}}` if that's possible?
> 
> Thanks :)

Done!

idnorton (approved):
Looks good, thanks :)

PR #940, opened 2021-08-16
Add telephone number validation to user-facing inputs

anorthall:
As per an earlier discussion in #emfcamp-web, user facing telephone input fields were not validated and the user could enter any value, even alphabetical characters.

I have modified the existing `TelField()` class to add basic validation. It will allow a `+` at the start of the input to permit international numbers, and beyond that only checks that the input is numeric (whitespace allowed) and between 10 and 15 characters.

I updated the volunteer sign up form to use `TelField()` instead of `StringField()`. I chose not to update the CfP admin portal to use `TelField()` as it seems that administrators should be able to do as they please. I couldn't find any other instances of telephone input besides the three listed in this paragraph.

### DIFF
--- a/apps/common/forms.py
+++ b/apps/common/forms.py
@@ -8,6 +8,7 @@ from wtforms.widgets.html5 import EmailInput
 from wtforms.compat import string_types
 from wtforms.widgets.core import html_params
 from email_validator import validate_email, EmailNotValidError
+import re
 
 
 class EmailField(StringField):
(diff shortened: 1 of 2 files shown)


apps/common/forms.py
```diff
@@ -56,6 +57,17 @@ class TelInput(Input):
...
 
+    def pre_validate(form, field):
+        if not re.search(r"^\+?[0-9 ]+$", form.data):
+            raise ValidationError(
+                "A telephone number may only contain numeric characters."
+            )
+
+        if not 9 < len(form.data) < 16:
+            raise ValidationError(
+                "A telephone number must be between 10 and 15 digits."
```
idnorton, line 68: Could this length limit be a problem for international numbers?

idnorton (approved):

PR #942, opened 2021-08-20
Basic implemention of in-house bar training (in-progress)

anorthall:
Issue ref #867.

I've been working on this a bit over the past few days (with help from @idnorton), and I won't have time to look at it again for a little while, so I thought it might be good to leave here for review. 

I've implemented an in-house mechanism for completing the bar training. It is fully working, tested and functional, although it is a bit rough around the edges at the moment. It:

- Imports questions and answers from JSON
- Imports training data from Markdown files
- Renders the data in order - educational material, then sections on that material, then the next educational material, etc.
- Highlights incorrect answers or unfilled answers
- Saves successful completion to the database
- Allows users to re-do the training for reference if they wish
- Prevents under 18s from accessing the training

Features that could be added:

- At the moment, all the content is displayed on one page. I don't actually think it looks *too* bad, but could be improved by paginating it in JavaScript.
- JavaScript validation could be added - although again the experience without it isn't terrible.
- UI could probably be generally improved even without pagination - front end isn't my thing.
- Implement a way to reference static material from the markdown files to enable some images to be included (or just host the images on a domain that won't change and include them as hardcoded HTML)

Welcome any feedback/comments/requested changes.

### DIFF
--- a/apps/volunteer/bar_training.py
+++ b/apps/volunteer/bar_training.py
@@ -1,143 +1,184 @@
-from urllib.parse import urlencode
-from hashlib import sha256
-import hmac
 import json
-
-from flask import request, current_app as app, abort, render_template
-from flask_login import current_user
-import requests
-
-from main import csrf, db, external_url
-from models import event_year
-from models.user import User
-from models.volunteer.role import Role
-from models.volunteer.volunteer import Volunteer
+import os
+import markdown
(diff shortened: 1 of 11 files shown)


marksteward:
Thanks for this! Looks perfect! Manually rebased and merged via #1050

PR #943, opened 2021-08-28
var/vat_invoices is required for Docker Compose

jellybob:
Without this directory existing `docker compose up` will fail to start the website container because its attempting to mount `var/vat_invoices`.


russss:
Any idea why this doesn't cause the integration tests to fail? I thought docker created nonexistent mountpoints.

jellybob:
Is it because the integration tests are overlaying `docker-compose.ci.yml`, which doesn't define those volumes? I can never remember whether multiple docker files add additional configuration to a container or completely override it.

marksteward:
To update in case anyone stumbles on this, it looks like `docker compose` (not `docker-compose`) is using bind mounts instead of volumes. To cope with missing directories, it creates them in the `docker compose create` step, but not in `docker compose start`. So you can reproduce this with:

```
docker compose create app
rmdir var/vat_invoices
docker compose start app
```

jellybob:
I should let Docker update more frequently. This is fine on the latest version of Docker Desktop for Mac.

Issue #950, opened 2021-09-27
Volunteer creation, shift signup


idnorton:
* Provide feature to create volunteers directly in the web UI
* Sign a volunteer to a shift

idnorton:
Simple attempt failed:
```
diff --git apps/volunteer/admin/volunteer.py apps/volunteer/admin/volunteer.py
index 3eac3481..daf2d2c7 100644
--- apps/volunteer/admin/volunteer.py
+++ apps/volunteer/admin/volunteer.py
@@ -5,7 +5,7 @@ from models.volunteer.volunteer import Volunteer
 
 
 class VolunteerUserModelView(VolunteerModelView):
-    can_create = False
+    can_create = True
     can_delete = False
     can_set_page_size = True
     can_view_details = True
```
Error:
```
Integrity error. (psycopg2.errors.NotNullViolation) null value in column "user_id" of relation "volunteer" violates not-null constraint DETAIL: Failing row contains (26, null, null, Ian, f, f, 1234, i.d.norton+test123@gmail.com, null, t, t). [SQL: INSERT INTO volunteer (planned_arrival, planned_departure, nickname, missing_shifts_opt_in, banned, volunteer_phone, volunteer_email, over_18, allow_comms_during_event, user_id) VALUES (%(planned_arrival)s, %(planned_departure)s, %(nickname)s, %(missing_shifts_opt_in)s, %(banned)s, %(volunteer_phone)s, %(volunteer_email)s, %(over_18)s, %(allow_comms_during_event)s, %(user_id)s) RETURNING volunteer.id] [parameters: {'planned_arrival': None, 'planned_departure': None, 'nickname': 'Ian', 'missing_shifts_opt_in': False, 'banned': False, 'volunteer_phone': '1234', 'volunteer_email': 'i.d.norton+test123@gmail.com', 'over_18': True, 'allow_comms_during_event': True, 'user_id': None}] (Background on this error at: http://sqlalche.me/e/13/gkpj)
```

idnorton:
Okay I thought I'd done this but quick testing to verify suggests that it's not working right. Still getting:
```
[...]

PR #956, opened 2021-09-28
Iann/950 allow volunteer team to create volunteers

idnorton:
As per #950


ormiret:
I think needs some more work to handle link to `User` (either creating a new one or linking to it if volunteer already has an account with main system). 

When I tried creating a `Volunteer` getting an error for the user link:
```
Integrity error. (psycopg2.errors.NotNullViolation) null value in column "user_id" of relation "volunteer" violates not-null constraint 
```

PR #961, opened 2021-10-19
Remove assets block in arrivals template

idnorton:
Replace with local copies of required JS libraries following deprecation of assets config.

Starter for discussion - I suspect that there will be opinions around the JS files being local or not and I'm not sure if they're in the correct place so please let me know if they should be somewhere else.


marksteward:
No issue in principle with vendoring these, but:
- they should go in a `vendor/` directory
- we could do with a test that tells us when they're out of date.

idnorton:
@marksteward is that the case for other things in that directory too?

russss:
On further consideration I think I'd rather see this built with Gulp like the rest - that means that our dependencies are in `package.json` and dependabot can monitor them.

I'm happy to sort this out.

idnorton:
@russss that sounds like a sensible plan. I wasn't sure the best way to take this forward so was very much looking for guidance.

russss:
I've had a stab at this in da7e726b. The arrivals page at least loads now, without JS errors.

It's not especially clear how to get hold of the datatables.searchhighlight plugin through NPM, so that might break stuff with actual data, but we can cross that bridge when we come to it. I think we can probably replace datatables here with a few lines of JS to be honest.

PR #965, opened 2021-11-13
Fix data: EMF 2016 did not last for two years

philandstuff:
This should fix [this page](https://www.emfcamp.org/schedule/2016).


marksteward (approved):
It was *technically* correct, but this is better 😉

Thanks!

PR #970, opened 2022-01-05
Update pywisetransfer to 0.1.9

marksteward:
pywisetransfer now supports SCA so we can fetch transactions in response to webhooks. See [`docs/wise.md`](https://github.com/emfcamp/Website/blob/dc4ac6d1/docs/wise.md).

Also adds a command `flask sync_wisetransfer` that downloads statements on demand in case webhooks fail.

### DIFF
--- a/apps/base/tasks_banking.py
+++ b/apps/base/tasks_banking.py
@@ -8,6 +8,11 @@ from sqlalchemy.orm.exc import NoResultFound
 from main import db
 from apps.base import base
 from apps.payments import banktransfer
+from apps.payments.wise import (
+    wise_retrieve_accounts,
+    wise_business_profile,
+    sync_wise_statement,
+)
 from models.payment import BankAccount, BankTransaction
 
 
@@ -147,6 +152,21 @@ def load_ofx(ofx_file):
     app.logger.info(
         "Import complete: %s new, %s duplicate, %s dubious", added, duplicate, dubious
     )
+
+
+@base.cli.command("sync_wisetransfer")
+@click.argument("profile_id", type=click.INT, required=False)
+def sync_wisetransfer(profile_id):
+    """Sync transactions from all accounts associated with a Wise profile"""
+    if profile_id is None:
+        profile_id = wise_business_profile()
+
+    tw_accounts = wise_retrieve_accounts(profile_id)
+    for tw_account in tw_accounts:
+        # Each sync is performed in a separate transaction
+        sync_wise_statement(
+            profile_id, tw_account.borderless_account_id, tw_account.currency
+        )
 
 
 @base.cli.command("reconcile")
--- a/apps/payments/wise.py
+++ b/apps/payments/wise.py
@@ -251,13 +251,8 @@ def _collect_bank_accounts(borderless_account):
         )
 
 
-def wise_retrieve_accounts():
-    business_profile = wise_business_profile()
-
-    if not business_profile:
-        return
-
-    borderless_accounts = wise.borderless_accounts.list(profile_id=business_profile)
+def wise_retrieve_accounts(profile_id):
+    borderless_accounts = wise.borderless_accounts.list(profile_id=profile_id)
 
     for borderless_account in borderless_accounts:
         for bank_account in _collect_bank_accounts(borderless_account):
--- a/apps/payments/wise.py
+++ b/apps/payments/wise.py
@@ -4,7 +4,6 @@ import logging
 from datetime import datetime, timedelta
 from flask import abort, current_app as app, request
 from pywisetransfer.webhooks import verify_signature
-from pywisetransfer.exceptions import WiseException
 
 from models.payment import BankAccount, BankTransaction
 from . import payments
@@ -44,6 +43,7 @@ def wise_webhook():
     valid_signature = verify_signature(
         request.data,
         request.headers["X-Signature"],
+        app.config["TRANSFERWISE_ENVIRONMENT"],
     )
     if not valid_signature:
         logger.exception("Error verifying Wise webhook signature")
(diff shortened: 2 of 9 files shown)


main.py
```diff
@@ -141,8 +142,12 @@ def load_user(userid):
     login_manager.anonymous_user = load_anonymous_user
 
     stripe.api_key = app.config["STRIPE_SECRET_KEY"]
-    pywisetransfer.environment = app.config["TRANSFERWISE_ENVIRONMENT"]
-    pywisetransfer.api_key = app.config["TRANSFERWISE_API_TOKEN"]
+    global wise
+    wise = pywisetransfer.Client(
+        api_key=app.config["TRANSFERWISE_API_TOKEN"],
+        environment=app.config["TRANSFERWISE_ENVIRONMENT"],
+        private_key_file=app.config.get("TRANSFERWISE_PRIVATE_KEY_FILE"),
```
jayaddison, line 149: Checking my understanding: this is a `get` instead of a direct index lookup so that the app can be configured to work in test/sandbox environments without a private key for SCA available?

main.py
```diff
@@ -141,8 +142,12 @@ def load_user(userid):
     login_manager.anonymous_user = load_anonymous_user
 
     stripe.api_key = app.config["STRIPE_SECRET_KEY"]
-    pywisetransfer.environment = app.config["TRANSFERWISE_ENVIRONMENT"]
-    pywisetransfer.api_key = app.config["TRANSFERWISE_API_TOKEN"]
+    global wise
```
jayaddison, line 145: Not at all a blocker, but this feels like a sign that the interface for `pywisetransfer` might not be great.  Use of globals feels non-ideal and if there's a way to avoid that (again) in future that'd be nice.

apps/base/tasks_banking.py
```diff
@@ -149,6 +154,21 @@ def load_ofx(ofx_file):
     )
 
 
+@base.cli.command("sync_wisetransfer")
+@click.argument("profile_id", type=click.INT, required=False)
+def sync_wisetransfer(profile_id):
+    """Sync transactions from all accounts associated with a Wise profile"""
+    if profile_id is None:
+        profile_id = wise_business_profile()
```
jayaddison, line 162: Again, just checking my understanding: Wise returns an arbitrary (single) business profile ID from their "list profiles" endpoint.

Is there a dangerous failure mode here where a user runs this sync command without a `profile_id` argument (since it's not required) and undefined results occur?

(there is an [exception raised](https://github.com/emfcamp/Website/blob/b39577dd4e164f3299ee9325d871f26bc5589409/apps/payments/wise.py#L198) when multiple profiles are returned by the endpoint, but it's unclear to me from the [comment](https://github.com/emfcamp/Website/blob/b39577dd4e164f3299ee9325d871f26bc5589409/apps/payments/wise.py#L191) whether that can happen)

apps/payments/wise.py
```diff
@@ -249,7 +228,7 @@ def _collect_bank_accounts(borderless_account):
 
         if account.bankDetails.currency == "GBP":
             # bankCode is the SWIFT code for non-GBP accounts.
-            sort_code = account.bankDetails.bankCode
+            sort_code = account.bankDetails.bankCode.replace("-", "")
```
jayaddison, line 231: Similar to the handling of `accountNumber`, it might be nice for `pywisetransfer` to try to hide some of the formatting details here (depending on the class of institution codes that `bankCode` is intended to represent; I'm not sure off the top of my head what that is)

main.py
marksteward: Using singleton globals here is a Flask pattern, although they also usually have a separate `init_app` that actually configures the modules. I'm unsure the answer is to go back to Stripe's pattern of storing state in the module, but perhaps it's worth adding a Flask wrapper specifically to support this pattern?

apps/base/tasks_banking.py
marksteward: This is safe, because `sync_wise_statement` only processes BankAccounts which are marked as active in the admin page. The command will also currently fail if we suddenly get new, unrecognised BankAccounts, but I see no reason we shouldn't just fetch all profiles. (A BankAccount corresponds to a balance in Wise terminology, which is tied to a borderless_account_id and currency. Only one BankAccount per currency will be marked as active currently.)

main.py
jayaddison: Yep, some kind of flask plugin is probably the way to go, middle-term.

That said: I think I've also seen some kind of app-level global storage on the Flask `app` object before?  (`app.g.<your-variable-name> = <value>`, or similar - https://flask.palletsprojects.com/en/2.0.x/api/#flask.g?)

Shorter-term: fine as-is, albeit maybe with a comment to indicate that it might not be ideal practice (unless we think it's fine, in which case don't bother).

main.py
marksteward: Yep, and the API_TOKEN is defaulted to an empty string in the config example, so I assume it's not intended to be optional.

apps/payments/wise.py
marksteward: That would be nice for the `accountNumber` issue (which I haven't encountered yet, I wonder if we can get in touch with someone about that). Returning sort codes without hyphens might be surprising for users, so we'd probably still need this code even if pywisetransfer did some normalisation.

apps/base/tasks_banking.py
marksteward: Also, I expect we'll set `TRANSFERWISE_PROFILE_ID` on production, so `wise_business_profile` will always return that.

jayaddison: Ok, understood.  Setting `TRANSFERWISE_PROFILE_ID` sounds important in that case, and I'm good with this :+1:

jayaddison (approved):

PR #985, opened 2022-01-18
Add non-commital guidance for EU travellers

jellybob:


marksteward:
Sounds a bit shifty, and I'm not sure we want a heading of Brexit on our site, this is just how the UK works now.

PR #994, opened 2022-02-03
Require admission ticket for purchasing other items

russss:
We add a "requires_admission_ticket" attribute, which defaults to True, indicating that the user must have an admission ticket (either in their basket or already purchased) in order to purchase the product.

This prevents users from accidentally buying a parking or live-in vehicle ticket without having an admission ticket.

If the user doesn't have an admission ticket, an error is shown on the payment page, preventing them from completing the purchase.

image


apps/tickets/pay.py
```diff
@@ -108,13 +108,30 @@ def pay(flow="main"):
 
     form.basket_total.data = basket.total
 
+    # Whether the user has an admission ticket in their basket or already purchased.
+    # FIXME: this is rather ugly
+    has_admission_ticket = any(
+        (
+            p.product.is_adult_ticket()
+            and p.state not in ("cancelled", "refunded", "reserved")
```
marksteward, line 116: I agree this is a fiddly, but on the plus side it's only needed in this part of the process. Maybe clearer to do `p.state in ("paid", "payment-pending")`? (states at https://github.com/emfcamp/Website/blob/main/models/purchase.py#L13-L17)

marksteward (approved):

PR #999, opened 2022-02-14
Add EV info

elsmorian:


templates/about/getting-there.html
```diff
@@ -20,6 +20,13 @@ <h3>By Car</h3>
 
 <p>Once you're near the site, you'll see yellow temporary road signs for EMF - follow these and ignore your sat nav if it tells you differently.</p>
 
+<h4>Electric Vehicles</h4>
+<p>There will be no provision for EV charging at the event this year due to both a substantial cost in fuel for the generators on-site, as well as a predicted increase in demand.</p>
```
jellybob, line 24:
```suggestion
<p>Owing to the increased price of providing electricity at EMF, and the increasing number of electric
    vehicles, we do not plan to provide generally-available electric vehicle charging at EMF 2022.
    We ask that you arrive at EMF with enough charge to leave at the end of the event.</p>
```

(Keeps it consistent with wording on the power page)

templates/about/power.html
```diff
@@ -42,10 +42,10 @@ <h4>Camper Vans &amp; Caravans</h4>
 <h4>Electric Vehicles</h4>
 <p>Owing to the increased price of providing electricity at EMF, and the increasing number of electric
     vehicles, we do not plan to provide generally-available electric vehicle charging at EMF 2022.
-    We ask that you arrive at EMF with enough charge to leave at the end of the event.
-</p>
+    We ask that you arrive at EMF with enough charge to leave at the end of the event.</p>
```
elsmorian, line 45: This was a bit of a messy edit but leaving it in as it's actually more consistent with the rest of the page

PR #1001, opened 2022-02-14
About page updates

jellybob:
Migrate the villages page from the wiki into /about/villages in preparation for restructuring the wiki.

I've also updated the internet page to list the default username/password as `outboundonly` - I'm happy to debate this, but I think to most people connecting to a event wifi and immediately being exposed to the entire internet is surprising to say the least.


russss:
Can you remove the changes to the internet page for the moment, our intention is to change the default SSID to outbound-only this year, and this page is not accessible yet anyway.

PR #1002, opened 2022-02-15
Allow villages to link to a website/wiki page

jellybob:
All links on public pages are tagged not to be followed by Google and friends, and you have to be a registered user to sign up, so I don't think SEO spamming will be a problem.


marksteward:
Can you add `__versioned__ = {}` to the class so we have a history of changes? Don't want URLs being swapped out sneakily.

jellybob:
Neat, done.

jellybob:
And now it'll even work.

marksteward (approved):

PR #1010, opened 2022-02-23
Verify Wise webhook signatures based on X-Signature-SHA256 request header

jayaddison:
- Upgrade to [`pywisetransfer` v0.2.0](https://github.com/jayaddison/pywisetransfer/releases/tag/v0.2.0) (that has switched signature verification to use SHA256 instead of SHA1 since v0.1.9)
- Read the Wise signatures from the incoming `X-Signature-SHA256` HTTP header

### DIFF
--- a/apps/payments/wise.py
+++ b/apps/payments/wise.py
@@ -34,7 +34,7 @@ def wise_webhook():
     )
 
     try:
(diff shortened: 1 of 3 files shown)


marksteward (approved):

marksteward:
🥳

Issue #869, opened 2020-02-16
Volunteer template errors `jinja2.exceptions.TemplateSyntaxError: Encountered unknown tag 'assets'.`


idnorton:
Volunteer shift at http://127.0.0.1:5000/volunteer/schedule/ has this issue.

It's down to an issue with `template/volunteer/base.html` and deprecation of `ASSET_URL`.

Need to find and fix all instances of `ASSET_URL`

idnorton:
Getting past the error in `volunteer/base.html` with the following diff:
```
diff --git templates/volunteer/base.html templates/volunteer/base.html
index 50223d5e..ed41bc40 100644
--- templates/volunteer/base.html
+++ templates/volunteer/base.html
@@ -5,8 +5,9 @@
         <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <title>{% block title %}{% endblock %}{% if self.title() %} - {% endif %}EMF Volunteering</title>
-        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
-        <link rel="stylesheet" href="{{ static_url_for('static', filename='css/volunteer.css') }}">
+        {% block css %}                ￼
+           <link rel="stylesheet" href="{{ static_url_for('static', filename="css/main.css") }}">
+        {% endblock %}
         {% block head -%}{% endblock -%}
         <link rel="shortcut icon" href="/favicon.ico">
     </head>
@@ -39,11 +40,7 @@
     </div>
 </div>
 {% endblock %}
-<script src="//code.jquery.com/jquery-1.10.1.min.js"></script>
-<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js" integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa" crossorigin="anonymous"></script>
-{% assets "js_main" -%}
-<script src="{{ ASSET_URL }}"></script>
-{% endassets -%}
+<script src="{{static_url_for('static', filename="js/main.js")}}"></script>
[...]

PR #1013, opened 2022-02-24
Migrate /notify to /volunteer/admin/notify

idnorton:
Move and update notify to new location prior to improving notify.

Ref #869


marksteward (changes requested):
I think this one should be something like `volunteer_notify`. The name is used as a prefix for handlers, so referencing the `notify_email` handler in other parts of the app would then be something like `url_for("volunteer_notify.notify_email")`.

idnorton:
> I think this one should be something like `volunteer_notify`. The name is used as a prefix for handlers, so referencing the `notify_email` handler in other parts of the app would then be something like `url_for("volunteer_notify.notify_email")`.

@marksteward are you saying that it shouldn't be located at the path it is or something else?

idnorton:
Based on the conversation we had on Thu Nov 04 2021 I think the intention is to remove the email sending aspects of this and use the shared libraries that the admin UI uses rather than having duplicate mechanisms, that can then take care of queueing and handing sending the messages as it does for the major admin notifications.

I'd like to restrict this change to a simple move and follow next steps with updating what;s there to better fit our needs rather than trying to do all of it in a single PR.

Much more a fan of lots of small PRs than a small number of huge PRs.

marksteward (approved):

PR #1052, opened 2022-03-26
Iann/volunteer notifications

idnorton:
* Deprecate legacy notify code
* Add replacement notify mechanisms via volunteer admin


idnorton:
Much of the code from here is copied from admin/email and switches approach to use VolunteerNotifyJob tables in order to send the messages *after* the person has written them instead of trying to do this immediately using SMTP send.

russss (approved):

PR #1057, opened 2022-03-26
Add VolunteerNotifyJob email handler!

idnorton:


apps/base/scheduled_tasks.py
```diff
@@ -3,6 +3,7 @@
 from main import mail, db
 
 from models.email import EmailJobRecipient
+from models.notify import VolunteerNotifyRecipient
```
marksteward, line 6: This should read `models.volunteer.notify`

PR #1058, opened 2022-03-26
Fix & persist volunteer schedule filters

jellybob:
All the filters work now, and those options are persisted across requests to local storage.

I've also removed the necessity to click a filter button, filters are immediately applied as options change now.


ormiret (approved):
Looks good

PR #1060, opened 2022-03-27
Diy proposal withdrawal

SamLR:
Allow users to withdraw their CfP proposals

### DIFF
--- a/apps/cfp/views.py
+++ b/apps/cfp/views.py
@@ -392,6 +392,36 @@ def edit_proposal(proposal_id):
     return render_template("cfp/edit.html", proposal=proposal, form=form)
 
 
+class WithdrawalForm(Form):
+    confirm_withdrawal = SubmitField("Confirm proposal withdrawal")
+
+
+@cfp.route("/cfp/proposals/<int:proposal_id>/withdraw", methods=["GET", "POST"])
+@feature_flag("CFP")
+def withdraw_proposal(proposal_id):
+    if current_user.is_anonymous:
+        return redirect(
+            url_for(
+                "users.login", next=url_for(".edit_proposal", proposal_id=proposal_id)
+            )
+        )
+
+    proposal = Proposal.query.get_or_404(proposal_id)
+    if proposal.user != current_user:
+        abort(404)
+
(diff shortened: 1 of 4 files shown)


apps/cfp/views.py
```diff
@@ -392,6 +392,49 @@ def edit_proposal(proposal_id):
...
+
+@cfp.route("/cfp/proposals/<int:proposal_id>/withdraw", methods=["GET", "POST"])
+@feature_flag("CFP")
+def withdraw_proposal(proposal_id):
+    if current_user.is_anonymous:
+        return redirect(
+            url_for(
+                "users.login", next=url_for(".edit_proposal", proposal_id=proposal_id)
+            )
+        )
```
marksteward, line 410: This could be @login_required I think

templates/cfp/proposals.html
```diff
@@ -12,8 +12,13 @@ <h2>Your proposals</h2>
   {% for proposal in proposals %}
   <tr class="{{- loop.cycle('odd', 'even') }}">
     <td>
-      <h4><a href="{{ url_for('.edit_proposal', proposal_id=proposal.id) }}">{{ proposal.title }}</a></h4>
-      <p>{{ proposal.human_type | title }}</p>
+      <h4><a href="{{ url_for('.edit_proposal', proposal_id=proposal.id) }}">{{ proposal.title }}{% if proposal.state in ('edit', 'new') %} [edit]{% endif %}</a></h4>
```
marksteward, line 15: This [edit] flag only makes sense to us. It's probably clearer to disable the fieldsets and Update button in `edit-form.html` if that's the intention (I'm guessing people should still be able to message about a withdrawn proposal).

marksteward (approved):

templates/cfp/proposals.html
SamLR: The intention on this was just to make it clear that they could click through on the title to edit their submission (because I don't think it's super obviously a URL)

apps/cfp/views.py
SamLR: good point

PR #1066, opened 2022-04-03
Build for arm64 as well as amd64

jellybob:
I'm not sure whether this actually works or not, but theoretically it should result in making arm64 images available for people running on recent Apple machines rather than having to translate from amd64 which sucks up battery life at an impressive rate.

Annoyingly this isn't a complete solution as PostGIS doesn't have ARM builds on Docker Hub, but it does get us most of the way there I believe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,17 @@ to monitor the containers.
 
 To start all containers (and rebuild any which may have changed):
 
+> [!TIP]
+> **Apple Silicon Users**
+> You'll probably want to build your own version of the base images as well, as
+> the ones we build are amd64 only, and so painfully slow on ARM devices. This
+> is unlikely to be resolved until GitHub Actions has ARM runners.
+>
+> ```sh
+> docker build -t ghcr.io/emfcamp/website-base -f ./docker/Dockerfile.base .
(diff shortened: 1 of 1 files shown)


marksteward:
I'm guessing this doubles the build time. Can we use a build matrix to target simultaneously?

I'm not sure about merging this if we know the postgres image won't work. Should this be a draft PR?

jellybob:
This might be doing a matrix build, assuming I've properly read the Github Actions docs. I've got no idea how to actually test it since builds only happen for `main`.

marksteward:
Looks right to me!

jayaddison:
> I've got no idea how to actually test it since builds only happen for `main`.

One important-ish note for testing: it's probably best to change the image tags from `latest` to `testing-1066` or something so that they don't risk being deployed unintentionally.

Because this branch/PR are opened from within `Website` repo, I _think_ you could modify the trigger conditions to `on: push, branch: build-arm64-images` or similar, and that should initiate a workflow run from this branch (just make sure to remember to revert that when satisfied with the testing results).

(I say this as someone who generally tests GHA changes by hackily making changes from a repository `main` branch :/ )

jellybob:
Closing until Github Actions has ARM runners, at which point I'll take another pass at this.

PR #1085, opened 2022-04-05
Charges don't have a succeeded state

marksteward:
It's too late rn to work out why this currently works in production. Maybe the API level is different, or maybe the events come in different orders.


marksteward:
https://stripe.com/docs/upgrades#2015-02-18

PR #1086, opened 2022-04-05
Update volunteer list search filter

idnorton:
Use interested roles not trained roles. We only have one role that
requires training so we're not getting to see who's actually volunteered
to do a particular role only those trained.

Casual drive by whilst we were triaging remaining volunteer issues last
night.


idnorton:
Never mind - @ormiret is being super efficient again and has it in https://github.com/emfcamp/Website/pull/1083/files ;)

PR #1095, opened 2022-04-22
Add form to change a proposal's owner

SamLR:
We often end up creating submissions for invited users. At some point
it's useful to change the ownership of these back to the correct people
so we can communicate with them via the site/keep them updated on the
status of their talk slot/send them speaker tickets

(or in some rare cases they just used the wrong email address etc).

This adds a new form which makes that change.

### DIFF
--- a/apps/cfp_review/base.py
+++ b/apps/cfp_review/base.py
@@ -49,6 +49,7 @@ from .forms import (
     AcceptanceForm,
     ConvertProposalForm,
     AddNoteForm,
+    ChangeProposalOwner,
 )
 from . import (
     cfp_review,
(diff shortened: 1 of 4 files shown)


russss (changes requested):
I feel like the "create new user" checkbox isn't really needed here - if the user exists then we can just ignore the name field. If the user doesn't exist and the name field isn't present then we can show an error.

We also deal with this issue on the `admin.tickets_issue_free` endpoint, which asks for the email first and then asks for the name on the next page if needed.

(Ideally it would be nice to have a reusable component to do this, perhaps with JS type-ahead-find, but that probably needs to go on the backlog.)

SamLR:
That look better, @russss?

russss (approved):

Issue #1113, opened 2022-05-05
Option to hide CfP items from line-up/schedule


russss:
Sometimes we want a CfP item to be in the "accepted" state (so that it's considered in scheduling) but not yet visible on the website (perhaps because it's dependent on funding etc).

I think the best way to do this would be with a "hide from schedule" flag, rather than a separate state.

PR #1115, opened 2022-05-05
Cfp updates

### LINKED ISSUE #1106: Make it obvious that there are notes on a CFP entry
Currently it's not clear that there are notes on CFP entries, add a "*" or a "(10)" or something to the tab bar to make this clearer. Perhaps also make it bold?

Ideally we should make the "details" tab show this in a non-editable form, along with all the other information that the content team need to manage it, and we move the big mess of textboxes to another tab?

SamLR:
closes #1106, #1113 and removes hard coded values for `DAYS`


models/cfp.py
```diff
@@ -206,6 +208,16 @@
 MANUAL_REVIEW_TYPES = ["youthworkshop", "performance", "installation"]
 
 
+# This is a function rather than a constant so we can lean on the configuration
+# for event start & end times rather than hard coding stuff
+def get_days_map():
+    event_days = [
+        (event_start() + timedelta(days=day_idx))
```
marksteward, line 215: I think this should explicitly call .date(), as event_start is currently 11am and we add hours on below

marksteward (changes requested):

marksteward:
I rebased and merged this already.

PR #1123, opened 2022-05-08
Attendee content

jellybob:
Provide an interface for attendees to add their own content to the schedule.

Venues can be associated with a village (only village members can add content to it), can be flagged as being for scheduled content only (only content from the CFP can go there), or neither (anyone can add content there, for places like the bar).

Content added via this path immediately jumps to a finalised state, and is flagged as being user scheduled. Still to do is making slotmachine respect that and not attempt to reschedule it or put it somewhere else.


Jonty:
We're only putting CFP-content in the villages at certain times and the rest of the time the village can do whatever they like. Does this currently work in that situation?

jellybob:
I've not tried it yet, but this should allow slotmachine to schedule CFP content at any times the village isn't already running something in their workshop. There's no validation to prevent a village member scheduling content in a workshop at the time as CFP content right now though.

Jonty:
I think it's ok for villages to do that as they might be running more than one thing at once, but we definitely need to warn them when this happens. And maybe us.

migrations/versions/7c8d52a17f8f_add_village_venues.py
```diff
@@ -0,0 +1,30 @@
...
+
+Revision ID: 7c8d52a17f8f
+Revises: 1012d60b8c68
+Create Date: 2022-05-07 20:47:38.189635
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '7c8d52a17f8f'
+down_revision = '1012d60b8c68'
```
marksteward, line 11: As I've just merged #1128 you'll need to change this to `822ce04a81a2`. Probably easiest to do `./flask db downgrade` twice, then rebase/merge, make the change and upgrade 3 times.

marksteward (changes requested):

marksteward (approved):

marksteward (approved):

Issue #794, opened 2018-09-07
Speaker checkin page


marksteward:
For green room and heralds. Also to include prompt for heralds with messages, timings and blurb. This might be worth splitting into a GR view and herald/VOC view, to help ensure opted-out talks aren't recorded.

> atm our issues are: not everyone shows up at GR; often they show up at the stage, so heralds should be able to check people in in a pinch; the check in stage should update the CFP flags for recording etc.; herald quality of intros were... mixed, so a single method would be great for those who are new to it; messages should be better/more easily communicated; timings too
> 
> having a site for heralds to get all the info about what they're doing, where they are, messages, etc. would be SUPER useful

SamLR:
I think this might be usefully handled on the 'per shift' view for volunteering. Ideally it should be a "here is all you need to know about this shift" style page with who to contact as well as method of flagging when someone's not turned up to relieve you.

Not sure if you agree @marksteward but I certainly agree that some sort of page for logging/displaying stuff whilst a talk is on-going is needed

dominicgs:
From the speakerops debrief:

* Ability to choose a stage
* Speaker check in (simple - speaker has arrived)
* Recording flags (and updating the CFP flags if they change their mind)
* who’s on now and how long they have to speak 
* Who’s up next (and when)
* Timing (not necessarily strict timing, as things may be late, perhaps a timer they can trigger manually that counts down for how long the talk is supposed to be)
* Messages to read out

dominicgs:
A similar page for workshops might be appropriate too, although I think we need some other tweaks to how workshops are handled - which I shall raise in a separate issue.

SamLR:
These now exist as the `/herald/<venue>` and `/greenroom` pages although they probably require refinement

PR #1128, opened 2022-05-10
Add 'published_pronouns' field to Proposals

### LINKED ISSUE #1124: Ask for pronouns in CfP finalisation
We've had instances of heralds using incorrect pronouns and upsetting speakers in the past, we should ask for preferred pronouns in finalisation and ensure the heralds have this information when introducing talks.

SamLR:
We want to enable proposers to list their pronouns. This adds the field
to finalisation forms, line up detail view and admin views.

This will support the herald page #794

closes #1124


templates/cfp/accepted.html
```diff
@@ -13,6 +13,9 @@
           collaborators to appear in the schedule. Any names you give here
           are separate to the one attached to your ticket.</p>
         {{ render_field(form.name, 9) }}
+        <p>What are your preferred pronouns? These will be displayed in the
```
marksteward, line 16: "What pronouns should we use?" maybe. Preferred pronouns is generally recommended against.

marksteward (changes requested):

Jonty:
Probably worth giving examples too, in case anyone isn't quite sure what we're asking

SamLR:
fixed both

PR #1136, opened 2022-05-14
Lightning talks

### LINKED ISSUE #1037: Add lightning talks as a CfP Proposal type
This will need to be another type. It will mostly operate for scheduling and planning so not used for e.g. full reviewing. 

Likely fields:

* title
* email (via login)
* short description 

These will be submit-able during the event (e.g. up until 2 hours before the session).

These will not make submitters legible to buy tickets.

SamLR:
closes #1037


SamLR:
Things to fix (still)

[ ] basic state ('withdrawn' should hide it)
[ ] ideally someway to modify the number of slots per session (or only release some)

models/cfp.py
```diff
@@ -792,6 +803,55 @@ class InstallationProposal(Proposal):
     funds = db.Column(db.String, nullable=True)
 
 
+class LightningTalkProposal(Proposal):
+    __mapper_args__ = {"polymorphic_identity": "lightning"}
+    human_type = HUMAN_CFP_TYPES["lightning"]
+    slide_link = db.Column(db.String, nullable=True)
+    # This should probably use allowed_times or something but that has a very
```
marksteward, line 810: I think this is the right way, so we can scratch this comment to prevent later doubt :)

models/cfp.py
```diff
@@ -792,6 +803,55 @@ class InstallationProposal(Proposal):
...
+            day: (
+                slots.get(day, (120 / LIGHTNING_TALK_LENGTH)) - day_counts.get(day, 0)
+            )
+            for day in LIGHTNING_TALK_SESSIONS.keys()
+        }
+
+    @classmethod
+    def get_total_lightning_talk_slots(cls):
+        sessions = TalkProposal.query.filter(
+            TalkProposal.title.startswith("Lightning Talk"),
```
marksteward, line 840: Feels like we should have a PlaceholderProposal or something, that inherits directly from TalkProposal. Can refactor before next time, though.

marksteward (approved):

models/cfp.py
SamLR: yea, in theory next time we could add a further constraint to Slot Machine that forces all lightning talks for a single day to be together (with no time between) then they're basically 'just' a talk but will have to do some thinking... later

PR #1161, opened 2022-05-22
Fine-grained CfP Closing

SamLR:
add per proposal type 'closed' flags and stop people signing up for lightning talk slots for today/the past


marksteward:
We probably want to distinguish "has opened" from "has closed" anyway.

marksteward (approved):

PR #1187, opened 2022-05-28
Volunteer sys

ormiret:


apps/volunteer/choose_roles.py
```diff
@@ -108,3 +120,62 @@ def role(role_id):
...
+    if current_user.is_authenticated:
+        if int(args[0]) in [
+            ra.role_id for ra in current_user.volunteer_admin_roles
+        ] or current_user.has_permission("volunteer:admin"):
+            return f(*args, **kwargs)
+        abort(404)
+    return app.login_manager.unauthorized()
+
+
+@volunteer.route("role/<role_id>/admin")
```
SamLR, line 145: I believe these parameters can have type hints (I think the format is `<int:role_id>` and, given they're going to be treated as ints by the `role_admin_required` decorator we should probably use them

apps/volunteer/choose_roles.py
```diff
@@ -108,3 +120,62 @@ def role(role_id):
...
+        Shift.query.filter_by(role=role)
+        .filter(Shift.end >= cutoff)
+        .order_by(Shift.end)
+        .limit(5)
+        .all()
+    )
+    return render_template("volunteer/role_admin.html", role=role, shifts=shifts)
+
+
+@volunteer.route("role/<role_id>/toggle_arrived/<shift_id>/<user_id>")
```
SamLR, line 160: [can be merged but we should fix later] I think this should be a POST not a get as it's changing state. Not sure what @russss & @marksteward's thoughts on this are though so happy to be over ruled and can probably merge despite it.

apps/volunteer/main.py
```diff
@@ -21,3 +26,78 @@ def safeguarding():
 @feature_flag("VOLUNTEERS_SIGNUP")
 def info():
     return render_template("volunteer/info.html")
+
+
+@volunteer.route("/init_shifts")
```
SamLR, line 75: [can be merged] This should probably be in a tasks file (e.g. https://github.com/emfcamp/Website/blob/main/apps/cfp/tasks.py)

apps/volunteer/main.py
```diff
@@ -21,3 +26,78 @@ def safeguarding():
...
+                for s in shifts:
+                    db.session.add(s)
+
+    db.session.commit()
+    return redirect(url_for(".main"))
+
+
+@volunteer.route("/clear_data")
+@v_admin_required
+def clear_data():
```
SamLR, line 133: likewise, tasks file

templates/volunteer/info.html
```diff
@@ -4,16 +4,41 @@
...
+   
     <h2>Important numbers</h2>
-    <dl class="dl-horizontal">
+    <!-- <dl class="dl-horizontal">
         <dt>First aid (mobiles)</dt><dd>01908 870909</dd>
         <dt>First aid (tent)</dt><dd>01908 870908</dd>
         <dt>Site manager</dt><dd>01908 870904</dd>
         <dt>Conduct</dt><dd>01908 870912</dd>
+    </dl> -->
+    <dl class="dl-horizontal">
```
SamLR, line 33: There needs to be something making it clear that these are ~DECT~ SIP numbers and what that means (e.g. that fixed locations will have a phone that can be used for these).

templates/volunteer/info_emergency.html
```diff
@@ -0,0 +1,31 @@
...
+  <p>If you encounter a fire (and it's not meant to be there), follow the steps below:</p>
+  <ol>
+    <li>Notify HQ via radio or phone (1011), making sure to clearly describe the location.</li>
+    <li>Ensure people are clear of danger.</li>
+    <li>If it is safe to do so, attempt to put out the fire.</li>
+  </ol>
+
+  <h3>First Aid</h3>
+
+  <p>In most cases people who need first aid should make their own way to the first aid tent, which can be found near the bar.</p>
```
SamLR, line 27: [cbm] map link as well?

templates/volunteer/info.html
```diff
@@ -4,16 +4,41 @@
...
+    <p><strong>Don’t overcommit yourself.</strong></p>
+    
+    <p><strong>Tell us if you can’t get something done.</strong></p>
 
+    <p>Nobody will criticise you if you can't do something, and telling us
+    quickly means we can handle the situation.</p>
+   
     <h2>Important numbers</h2>
-    <dl class="dl-horizontal">
+    <!-- <dl class="dl-horizontal">
```
SamLR, line 27: [nitpick] commented out stuff to remove again

SamLR (approved):
I think this looks ok to me, some minor tidy up tasks.

apps/volunteer/main.py
ormiret: Went with view rather than CLI task to trigger this as I can hit the URL for it but don't have access to run CLI tasks.

Issue #1167, opened 2022-05-24
redirect 'foo/' paths to 'foo'


SamLR:
Mostly because I keep trying to go to '/cfp-reviews/' which is a 404 and I'm lazy (and I want computers to fix that for me)

marksteward:
Fixed by #1189

PR #1189, opened 2022-05-28
Allow trailing slash in cfp-review and arrivals

kragniz:
So people don't get a 404 if they leave the slash in the url.

### DIFF
--- a/apps/arrivals.py
+++ b/apps/arrivals.py
@@ -28,7 +28,7 @@ arrivals_required = require_permission(
 )  # Decorator to require arrivals permission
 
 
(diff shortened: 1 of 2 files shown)


kragniz:
Somewhat related to #1167.

SamLR:
I think this can be solved more generally using this https://stackoverflow.com/questions/40365390/trailing-slash-in-flask-route but I've not investigated it yet

marksteward:
Don't want to add this generically, it's only meaningful on certain URLs. I'd also prefer to normalise URLs rather than silently ignore the trailing slash.

Issue #1217, opened 2022-06-01
Saving to home screen from iOS does not work


edwardhorsford:
System: iOS 15.4.4

## What I was doing:
Attempted to save the emf schedule to my home screen from IOS

## What happened:
When opening the saved site, it 404s

## What I can see
The url that saves is https://www.emfcamp.corg/static/ - which 404s.

I *think* this is chosen because these things use the `manifest.json` and it sets the `site_url` to the relative `.`. My guess is that as the manifest loads off of the static directory, the url suggested for the saved site is thus static.

I have half a hunch that changing the `site_url` to `/` might fix things - or possibly removing it? I wonder if removing it would save the url that the user performed the action? this would be better for me as I wanted to save the schedule.

edwardhorsford:
Fixed by #1218

PR #1218, opened 2022-06-01
Remove start_url which causes 404 url to be saved

edwardhorsford:
Per #1217, this relative url seems to cause iOS to save the static directory to the home screen - which does not work.

This _could_ be set to the root directory. But [by removing this key](https://stackoverflow.com/questions/39495573/set-start-url-in-web-app-manifest-based-on-the-url-that-was-added-to-homescreen), it will save whatever page the user was on - in my use case this would let me save the schedule to my home screen.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -4,7 +4,6 @@
   "name": "Electromagnetic Field",
   "short_name": "EMF",
   "lang": "en-GB",
-  "start_url": ".",
   "display": "standalone",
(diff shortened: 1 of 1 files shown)


edwardhorsford:
NB: I don't have the app set up, so am unable to test this. It's my best guess that this is the cause of #1217

marksteward:
Confirmed, thanks!

PR #1255, opened 2022-06-30
Add role admin view of all volunteers for that role

ormiret:


idnorton (approved):
Looks good to me, though I haven't tested it.

PR #1261, opened 2022-10-12
Updated README.me for Red Hat users

fragmad:
There is a potential gotcha when attempting to run the docker-compose within Red Hat based systems where SELinux interferes with the starting of the app and gulp containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,13 @@ try:
 
         docker compose up --build
 
+### Errors starting the app and gulp containers 
+
+In RedHat based systems e.g. Red Hat Enterprise Linux, Fedora Core, etc you may need to set SELinux to Permissive for the Entrypoint script and other bash scriptsto have the correct permissions. Try: 
+
+        sudo getenforce
+        sudo setenforce permissive
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,13 @@ try:
 
         docker compose up --build
 
+### Errors starting the app and gulp containers 
+
+In RedHat based systems e.g. Red Hat Enterprise Linux, Fedora Core, etc you may need to set SELinux to Permissive for the Entrypoint script and other bash scriptsto have the correct permissions. Try: 
+
+        sudo getenforce
+        sudo setenforce permissive
```
idnorton, line 55: Looks like we've got some inconsistent styling in this file but could you outdent this to the first column and quote it as a command block please?

Thanks for the contribution!

marksteward:
I don't think we should be encouraging people to disable SELinux.

What method did you use to run docker? I've followed https://docs.docker.com/engine/install/centos/ on an enforcing RHEL 8 and it just worked.

russss:
Closing stale issue

PR #1274, opened 2023-02-26
Add data export for volunteer summary stats.

ormiret:


russss:
If this is intended for the post-event export then you should implement the `get_export_data` class method on the model class, e.g.:

https://github.com/emfcamp/Website/blob/82f266ddcf9eb05c86d7c1f4d160ccea7f10c0ff/models/user.py#L273-L288

The public data will get committed to the repo when we run the export, and any private data will be stored until the next event.

russss (approved):

Issue #878, opened 2020-02-19
SEPA payment references should be ISO 11649 Creditor References


davidc:
SEPA payment references should be amended to be ISO 11649 Creditor References, with "RF" and with two check digits added to the start.

This will allow the user's bank to verify the reference prior to the payment being sent to avoid typos.

Needs amending anywhere references are displayed (not necessary to change how they are stored in the database) plus any bank recon needs to accept both formats.

russss:
We can't do this while using Barclays for Euro banking as they severely truncate references leaving us as few as 2 characters to work with. So this depends at least on #857.

davidc:
This should not apply when giving BACS transfer details as it will run into reference length issues (and probably wouldn't be validated anyway)

jayaddison:
See https://wiki.xmldation.com/General_Information/ISO11649_-_Creditor_Reference for some details on the check digit algorithm.

There are a few libraries in various languages [on github](https://github.com/search?q=ISO+11649&type=Repositories) which may be able to validate/generate this format.

jayaddison:
> There are a few libraries in various languages [on github](https://github.com/search?q=ISO+11649&type=Repositories) which may be able to validate/generate this format.

For future reference - [python-stdnum](https://arthurdejong.org/python-stdnum/) seems to be a good library for calculating and validating check digits of all kinds, including ISO 11649.

jayaddison:
I've begun some work for this, generating ISO-11649 creditor references that are 16 characters in length total (including the `RFXX` prefix, and then a 12-character reference part).

The draft implementation is naive, and the reference part currently consists of (all base-10) the event year number YYYY followed by a zero-padded 8-digit order number.

I'm not sure that format is ideal, and as follow-ups I'm considering:

* Reversibly encode the existing `BankPayment.bankref` instead of the `Payment.id` - this should be backwards-compatible with existing bank transaction reconcilation if done correctly.
* Consider using base36 encoding for the reference part, allowing for more concise creditor references.
* Consider not zero-padding the reference part, again providing more concise creditor references (ref: [Finland Finance intro to structured creditor references](https://www.finanssiala.fi/wp-content/uploads/2021/03/Introduction_of_the_global_Structured_Creditor_Reference.pdf#page=10)).
* Add a plaintext-readable (non-QR-encoded) copy of the structured reference on the invoice and/or on an edge of the QR code to provide an additional manual verification check (ref [this online EPC QR generator](https://epc-qr.eu/?form) which does this).

PR #1287, opened 2023-12-16
Prep for #878: move invoice-related QR code from receipt modules to invoice modules

jayaddison:
These invoice-related QR methods were placed into receipt-related modules out of a mixture of laziness and wanting to reduce the number of code lines modified during #891.  This cleanup (as preparation for #878) relocates them to make their purpose and usage clearer.

(note: the skipped test _does_ run and fails as expected if the `pytest.mark.skip` is removed)


tests/test_invoice.py
```diff
@@ -0,0 +1,34 @@
...
+    qr_image = render_svg(qr_inline)
+
+    expected = [
+        "BCD",
+        "002",
+        "1",
+        "SCT",
+        "",
+        "Electromagnetic Field Ltd",
+        "GB21BARC20716472954433",
```
russss, line 25: These are our live bank details, which we shouldn't have in git. (It's a minor thing but people do occasionally harvest them for direct debit fraud.)

The test Euro IBAN is this one:

https://github.com/emfcamp/Website/blob/e8276d3d7f32fcaf9dd763cbe4d91a0f22a0fe0e/apps/base/tasks_banking.py#L43

russss (changes requested):

russss (approved):

PR #1291, opened 2023-12-17
Static content as markdown

jellybob:
This allows us to easily take content put together on pads and turn it into web content, without a tedious step of translating to EMF flavoured HTML.


Jonty:
This is great. Love that you casually rewrote all the pages in markdown without even calling it out ♥️

templates/_nav.html
marksteward, line 1: typo for volunteering

Jonty:
Shouldn't the upstream branch be `main` rather than `devcontainer` on this PR?

jellybob:
I'm upstreaming it on `devcontainer` for now because I didn't want to muddy things with a bunch of VSCode config. Once that's merged this'll rebase to `main`.

russss (changes requested):
A few yaks in the markdown loading logic:

Firstly, I'm not sure if you can always assume that the current working directory is the root of the flask app and I don't want to find out the hard way when that isn't the case. It took me way too long to work this out, but `os.path.join(app.root_path, app.template_folder)` appears to be the canonical way of getting the template folder path in flask.

Secondly, the potential path traversal issues here make me (and CodeQL) slightly uneasy. Flask won't pass us any URL components with a forwardslash in, so it's _probably_ not an exploitable path traversal on our current setup, but I wonder if it could be exploitable on Windows. There's also the risk that `render_markdown` gets re-used in a situation where the input isn't sanitised. We should defend against path traversal here.

I believe the best way of handling this these days is something along the lines of:

```python
from pathlib import Path
...
if not Path(markdown_file_path).resolve().is_relative_to(Path(template_dir).resolve()):
	return abort(404)
```

Who knows if that will satisfy CodeQL.

(pathlib is great and it could be used for more of these path operations, although flask doesn't use it for its paths internally.)

jellybob:
I think this is sorted now, CodeQL is still unhappy about it but I've poked at trying to force a path traversal and have been unable to do so.

russss (approved):

PR #1292, opened 2023-12-17
WIP: Add structured ISO-11649 creditor references for SEPA QR codes

### LINKED ISSUE #878: SEPA payment references should be ISO 11649 Creditor References

jayaddison:
The lack of structured ISO-11649 creditor references seems to be at least one potential/likely reason why many SEPA-QR enabled banking applications failed to parse our QR codes during testing in #891.

There are a few more todos here:

- [ ] Determine whether to use `Payment.id` (sequence) or `BankPayment.bankref` to derive the creditor reference.
- [ ] Ensure that this works correctly end-to-end including reconcilation of transactions after receiving Wise balance credit webhooks.
- [ ] Either here or in a follow-up PR: adding the amount, recipient name, creditor reference and IBAN in plaintext on the edges of the SEPA QR for additional human verification and reassurance purposes (see [this online SEPA QR generator](https://epc-qr.eu/?form) for a nice example).

Resolves #878.

### DIFF
--- a/models/payment.py
+++ b/models/payment.py
@@ -323,7 +323,10 @@ class BankPayment(Payment):
         if self.currency == "EUR":
             order_id = f"{event_year()}{self.id:08d}"
             order_check_digits = mod_97_10.calc_check_digits(order_id)
-            return f"RF{order_check_digits}{order_id}"
+            assert len(order_check_digits) == 2
+            customer_reference =  f"RF{order_check_digits}{order_id}"
+            assert iso11649.validate(customer_reference)
+            return customer_reference
         else:
             return self.bankref
 
--- a/models/payment.py
+++ b/models/payment.py
@@ -322,8 +322,7 @@ class BankPayment(Payment):
         # Derive an ISO-11649 payment reference for EUR-currency payments
         if self.currency == "EUR":
             order_id = f"{event_year()}{self.id:08d}"
-            order_check_digits = mod_97_10.calc_check_digits(order_id)
-            assert len(order_check_digits) == 2
+            order_check_digits = mod_97_10.calc_check_digits(f"{order_id}RF")
             customer_reference =  f"RF{order_check_digits}{order_id}"
             assert iso11649.validate(customer_reference)
             return customer_reference
--- a/models/payment.py
+++ b/models/payment.py
@@ -7,6 +7,8 @@ from sqlalchemy import event, func, column
 from sqlalchemy.orm import Session, aliased
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from sqlalchemy_continuum.utils import version_class, transaction_class
+from stdnum import iso11649
+from stdnum.iso7064 import mod_97_10
 
 from main import db
 from . import (
@@ -309,6 +311,21 @@ class BankPayment(Payment):
                 return BankAccount.query.filter_by(currency=currency, active=True).one()
             except (MultipleResultsFound, NoResultFound):
                 continue
+
+    @property
+    def customer_reference(self):
+        assert self.id, (
+            "Customer references can only be generated for payments that have been stored "
+            "in the database."
+        )
+
+        # Derive an ISO-11649 payment reference for EUR-currency payments
+        if self.currency == "EUR":
+            order_id = f"{event_year()}{self.id:08d}"
+            order_check_digits = mod_97_10.calc_check_digits(order_id)
+            return f"RF{order_check_digits}{order_id}"
+        else:
+            return self.bankref
 
 
 class BankAccount(BaseModel):
--- a/apps/payments/invoice.py
+++ b/apps/payments/invoice.py
@@ -39,7 +39,7 @@ def make_epc_qrfile(payment, **kwargs):
         name="EMF Festivals Ltd",
         iban=payment.recommended_destination.iban,
         amount=payment.amount,
-        reference=payment.bankref,
+        reference=payment.customer_reference,
         encoding=1,
     )
     qr.save(qrfile, **kwargs)
--- a/poetry.lock
+++ b/poetry.lock
@@ -2940,6 +2940,23 @@ files = [
 six = ">=1.4.0"
 
 [[package]]
+name = "python-stdnum"
+version = "1.19"
+description = "Python module to handle standardized numbers and codes"
+category = "main"
+optional = false
+python-versions = "*"
+files = [
+    {file = "python-stdnum-1.19.tar.gz", hash = "sha256:133ec82f56390ea74c190569e98f2fb14b869808b1d54785708f22d0fead8b3f"},
+    {file = "python_stdnum-1.19-py2.py3-none-any.whl", hash = "sha256:1b5b401ad3f45b798b0317313b781a433f5d7a5ff2c9feb8054664f76f78644e"},
+]
+
+[package.extras]
+soap = ["zeep"]
+soap-alt = ["suds"]
+soap-fallback = ["PySimpleSOAP"]
+
+[[package]]
 name = "pytz"
 version = "2023.3.post1"
 description = "World timezone definitions, modern and historical"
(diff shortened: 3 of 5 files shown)


jayaddison:
> Either here or in a follow-up PR: adding the amount, recipient name, creditor reference and IBAN in plaintext on the edges of the SEPA QR for additional human verification and reassurance purposes

It seems that most of the relevant details (IBAN, recipient name, reference) are displayed in close proximity to the EPC QR already -- so maybe we don't need this.

However we should display the full creditor reference in relevant invoices -- not only the shorter 8-char `bankref` value.

The [formatting helper](https://github.com/emfcamp/Website/blob/3e9fc57f3128d94a4e8b6040920aad046aedf401/apps/common/__init__.py#L71-L73) that we [use when displaying bankrefs in invoices](https://github.com/emfcamp/Website/blob/3e9fc57f3128d94a4e8b6040920aad046aedf401/templates/payments/invoice.html#L184) would reformat, for example, a creditor reference of `RF67E9HKFVR8Q` as `RF67-E9HKFVR8Q` - not a terrible display format, but perhaps we could improve upon that.

jayaddison:
@marksteward I'm trying to think how to ensure that transaction matching is able to handle structured creditor references robustly if we introduce them, and one quirk of the current `safechar` set is that it includes both `R` and `F` (the two ISO11649 reference prefix characters).

I think it might be safer and easier to write any code that matches/removes `RF[0-9][0-9]` structured reference + check digit prefixes if at least one of those two characters could be removed from the `safechars` list.

As a demonstration of possible edge case:

```pyi
>>> from stdnum import iso11649
>>> iso11649.is_valid('RF46RF23')
True
```

Any thoughts on updating `safechars` to switch `F` for `E`?  (my guess is that those two are considered confusable, and/or `e` is considered confusable with `a` - I don't see any other ambiguous cases at the moment for `E`, but perhaps I'm missing one)

marksteward:
FWIW, I've played around with generating EPC codes at https://www.qr-code-generator.com/solutions/epc-qr-code/ and the Revolut app ignores the IBAN for GB, because it switches the form to sort code/account number mode.

I'm a bit concerned this might be why other apps aren't processing the QR codes, and we might be setting ourselves up for more of a headache than it's worth.

russss:
We don't use Barclays any more, and at any rate SEPA references have significantly longer field limit than BACS references, so length isn't an issue.

jayaddison:
> I feel like we should keep bankref separate from creditor reference, and change the label to something like "reference for manual entry". We could show the creditor reference for EUR payments, to prevent people wondering where it is, but I don't see why it shouldn't basically be the invoice number.

That's what I thought originally, too -- but then I figured that the creditor reference format is a superset of `bankref` -- so it essentially boils down to adding the prefix and check digits.  I'm still mulling this over.

> FWIW, I've played around with generating EPC codes at https://www.qr-code-generator.com/solutions/epc-qr-code/ and the Revolut app ignores the IBAN for GB, because it switches the form to sort code/account number mode.

> I'm a bit concerned this might be why other apps aren't processing the QR codes, and we might be setting ourselves up for more of a headache than it's worth.

It seems that the test IBAN we're using (`GB47 ...`) doesn't pass the relevant ISO check-digit test, perhaps another reason for those scan failures.  Using `GB15` instead does work -- or we could use some other test IBANs (@russss mentioned [these](https://www.iban.com/testibans) and I've also found a few [here](https://iban.co.uk/examples.html)).  I'll adjust the values in the codebase to something that validates but is also clearly not a valid bank and is a test account soon.

jayaddison:
Here's an attempted EPC QR generation from some local changes in-progress (not pushed yet because I've duplicated a bunch of the code from `invoice.html` to `transfer-waiting.html`):

image

jayaddison:
> `safechars` (which is the Microsoft product key character set) doesn't have any vowels to reduce the chance of coming up with pronounceable words. But the reason it's short and dense is to allow manual entry and to round-trip through old bank systems that don't support separate references. Do we still have that problem with ISO-11649? Do Barclays still end up concatenating the account name and reference?

It sounds like the length limits are no longer a problem.  I'm hesitant to make larger changes though, given that the existing code is known-working.

The no-vowels argument also seems convincing to me - and given that it's based on a particular character set, it'd be nice to maintain that.

So roughly speaking I think I'm OK with tightening the pattern-matching code slightly.  One idea I've just had is to detect whether the entire `ref` is a valid ISO11649 value, and if so, to `compact` it (using `python-stdnum`) and then extract everything from char-5 onwards.

jayaddison:
> If that's going to lead to more confusion (I don't know what the various EU bank payment forms might look like), we can just exclude R and F from the start of the bankref, i.e. change [this](https://github.com/emfcamp/Website/blob/main/models/payment.py#L292) to:

I'm reluctant to change the generating-side code; it would probably be fine, but feels a bit wrong somehow.

I think that if the consumer-side code (the transaction matching) is careful only to remove `RF[0-9][0-9]` when it is followed by 8 `safechars` (allowing again for punctuation) then that would handle it.  Even if `RF23`, for example, were generated as the first few characters of a `bankref`, then that treatment should be safe regardless of whether the ISO11649 prefix and check digits are present.

jayaddison:
A recording of the recently-added EPC QR explainer tooltip in action:

[epc-qr-explainer.webm](https://github.com/emfcamp/Website/assets/55152140/ff8edb74-f120-4e05-980e-ad6405f7d58b)

russss:
We had a report that our payment reference didn't work when sending a payment from a Dutch bank, because it was expecting a reference in a standard format. (I think it's likely that there was a separate description field where our reference would have worked, but this proved a useful test case.)

I manually generated an ISO-11649 reference (`RF82FTHGBVX2`) and the payer successfully used that. We just got the payment and the incoming reference received in Wise was `RF82FTHGBVX2/20240221220607<IBAN>` (where `<IBAN>` is the payer's IBAN).

So that has conveniently validated whether this works on the banking side - although the reference does come through in a slightly interesting format, which wasn't auto-reconciled currently.

I was previously a little concerned about how this would work in practice, but I'm fairly happy now that we should move to using ISO-11649 references, at least for Euro payments.

jayaddison:
Thanks @russss - that's great to have confirmed.

To handle the additional reference-format encountered: perhaps we could check whether the `ref` up-to-a-`/`-delimiter is a valid ISO11649 reference (using `python-stdnum`), and if so omit the remainder of the string from the matching process?

russss:
I wouldn't guarantee that the format will always be the same, although I'd hope it's the case, so I think we should still assume that it can appear anywhere in the reference field.

I have a few more comments which I'll put inline, but I'm inclined to get this merged soonish now.

models/payment.py
```diff
@@ -310,6 +312,23 @@ def recommended_destination(self):
...
+            raise Exception(
+                "Customer references can only be generated for payments that have "
+                "been persisted to the database."
+            )
+
+        # Derive an ISO-11649 payment reference for EUR-currency payments
+        if self.currency == "EUR":
+            order_check_digits = mod_97_10.calc_check_digits(f"{self.bankref}RF")
+            customer_reference = f"RF{order_check_digits}{self.bankref}"
+            assert iso11649.validate(customer_reference)
```
russss, line 327: This should be `iso11649.is_valid` I think

russss (changes requested):

models/payment.py
jayaddison: Perhaps it would also make sense to move that onto a `@validates` method on the SQLAlchemy model?

jayaddison: Nope, never mind: the structured reference isn't persisted (...yet?).  This could potentially be added in future.

Issue #678, opened 2018-06-08
Proposal topics/tags


russss:
We want to assign topics to each proposal, for monitoring purposes and maybe use in the schedule? These will not be used/visible in the review interface.

There will be a fixed list of topics, and multiple topics can be applied to the same proposal by admins.

Jonty:
Additionally:

* We need to be able to filter proposals by tag/state
* We need a view which shows (by state) the tags by %

It would be nice to pre-define the approved tags up-front to ensure we don't end up with deviation.

Stretch goals:
* Setting % goals on a per-tag basis, and showing which tags are over/under those percentages on the monitoring interface

SamLR:
closed by #1295

PR #1295, opened 2023-12-22
Add CfP tags

SamLR:
This adds tags to the CfP. 

Currently they're added via a basic text box in the cfp admin update-proposal view. 

Later PRs will add some nicer auto-complete/suggest functionality as well as view-by/filter-by tag tech.


apps/cfp_review/base.py
```diff
@@ -311,6 +312,13 @@ def log_and_close(msg, next_page, proposal_id=None):
                 next_id=next_id,
             )
 
+        # Create any new tags that are needed
+        new_tags = [t for t in prop.tags if t.id is None]
+        if any(new_tags):
+            for t in new_tags:
+                db.session.add(t)
+            db.session.commit()
+
```
marksteward, line 321: Could you add a comment that this is done in a separate transaction to prevent races against other sessions, maybe pull it out into a separate function called something like `ensure_tags` that does an upsert? I can see that we don't want to blow up after sending emails.

models/cfp_tag.py
```diff
@@ -0,0 +1,48 @@
...
+        return ",".join([str(t) for t in tag_list])
+
+    @classmethod
+    def parse_serialised_tags(cls, tag_str: str) -> list["Tag"]:
+        res = []
+        tag_list = [t.strip().lower() for t in tag_str.split(",")]
+        for tag_value in tag_list:
+            if len(tag_value) == 0:
+                continue
+            tag = cls.query.filter_by(tag=tag_value).one_or_none()
```
marksteward, line 33: Maybe this could be moved to the `ensure_tags` function too?

SamLR:
see https://github.com/emfcamp/Website/issues/678

apps/cfp_review/base.py
SamLR: Realised that this is no longer needed. I'd added this as a fix for the uniqueness error I was seeing from `proposal_tag_version` when adding 2 new tags but that was actually because I wasn't defining a compound key to the `proposal_tag` table (so the `_version` was only paying attention to one of the `ids`)

models/cfp_tag.py
SamLR: I don't think this function is needed now

SamLR:
I've tried (very lazily) to tame the proposal update form by putting the whole thing into an accordion. It's really shitty UI but it at least means you can navigate a little quicker. Might be possible to improve it by replacing the heading with summary of any set values in the section but that's a bit more than I feel like implementing right now.

@Jonty any thoughts as you're one of the folks who have interacted with this a lot?

Screenshot 2023-12-24 at 17 56 03

marksteward (approved):

SamLR:
Thanks @marksteward

PR #1296, opened 2024-01-04
Merged Data from YouTube

MattGrayYes:
Added YouTube URL and also original video filename.


marksteward:
I don't think this is the correct format, but don't worry, I think I can fix it.

MattGrayYes:
OK, I based it on 2018
https://github.com/emfcamp/Website/blob/main/exports/2018/public/schedule.json

marksteward:
2018 was also done manually, so I guess this is the new format.

I've run jq on the file so the diff shows up clearly - https://github.com/emfcamp/Website/compare/ddd6938..youtube-data. Thanks for this!

Issue #831, opened 2019-11-22
Improve talk video linkage


russss:
We want to have videos available on our official schedule pages as soon as they're published.

We have an [existing script](https://github.com/emfcamp/Website/blob/master/apps/base/tasks_videos.py) to link Youtube videos to talks in the schedule. This needs adding as a scheduled task and ensuring it actually works in 2020. 

We also need to link the videos on media.ccc.de, as these are more important - we don't store those at the moment. c3voc's API is [much simpler](https://api.media.ccc.de/public/conferences/emf2018). We need to extract the preview image link (`poster_url` from media.ccc.de) as it's useful for social media previews.

Lastly, we also need to make sure these are saved in the exported schedule.json as they now exist for previous years.

marksteward:
Noted in #1297 but not here: we use youtube_url internally, but not for the exports. The exports also aren't pretty printed but should be, and kept in sync with the .frab files (which CCC want for their ingester).

russss:
I don't think the frab files are really an issue. We aim to have the videos uploaded to youtube and media.ccc.de long before that frab export is created.

russss:
See also #1258

PR #1305, opened 2024-01-13
Allow (manually) populating video URLs mid-event

lukegb:
At the moment this is only accessible to CfP admins; maybe we should add a separate role/form for people who are allowed to attach uploaded videos to (accepted) proposals?

This also makes the talk URL visible in the frab XML (which will hopefully mean that media.ccc.de will link back to the talk's page) and updates a few bits to bring things into sync with c3voc's latest copy of the schema.

### DIFF
--- a/apps/cfp_review/base.py
+++ b/apps/cfp_review/base.py
@@ -391,6 +391,9 @@ def update_proposal(proposal_id):
     form.available_times.data = prop.available_times
     form.content_note.data = prop.content_note
     form.family_friendly.data = prop.family_friendly
+    form.thumbnail_url.data = prop.thumbnail_url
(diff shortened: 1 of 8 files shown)


marksteward:
Amazing, updating the frab schema is a bunch of fiddly stuff I was dreading having to do!

I think we can probably manage without a separate role, but we'll probably add a bunch more granular permissions at some point.

russss:
For the record, this fixes part of #831

lukegb:
I think this is actually ready for review; I think it remains to be seen how we'll actually end up pushing video URLs but this at least allows someone to yank/correct them manually if things go wrong.

marksteward (approved):

PR #1308, opened 2024-01-15
Add a 'delete_tags' cfp task

SamLR:
Because these will change and this way we can clean up

### DIFF
--- a/apps/cfp/tasks.py
+++ b/apps/cfp/tasks.py
@@ -134,7 +134,7 @@ def email_reserve():
 )
 @click.argument("tags_to_create", nargs=-1)
 def create_tags(tags_to_create):
-    """Upset tag list"""
+    """Upsert tag list"""
     if not tags_to_create:
         tags_to_create = DEFAULT_TAGS
 
@@ -150,3 +150,31 @@ def create_tags(tags_to_create):
 
     db.session.commit()
     app.logger.info(f"Successfully created {tags_created} new tags.")
+
+
+@cfp.cli.command(
+    "delete_tags",
+    help=f"Delete tags from the Database. Tagged proposals will have the tags removed.",
+)
+@click.argument("tags_to_delete", required=True, nargs=-1)
+def delete_tags(tags_to_delete):
+    """Delete tag list"""
+    tags_deleted = 0
+    for tag_name in tags_to_delete:
+        tag = Tag.query.filter_by(tag=tag_name).one_or_none()
+        if not tag:
+            app.logger.info(f"Couldn't find tag: '{tag_name}' exiting.")
+            return
+
+        if tag.proposals:
+            tagged_proposals = [p.id for p in tag.proposals]
+            app.logger.info(
+                f"'{tag_name}' will be removed from the proposals with ids: {tagged_proposals}"
+            )
+
+        db.session.delete(tag)
+        tags_deleted += 1
+        app.logger.info(f"'{tag_name}' added to session.")
+
+    db.session.commit()
+    app.logger.info(f"Successfully deleted {tags_deleted} tags.")


apps/cfp/tasks.py
```diff
@@ -150,3 +150,31 @@ def create_tags(tags_to_create):
...
     db.session.commit()
     app.logger.info(f"Successfully created {tags_created} new tags.")
+
+
+@cfp.cli.command(
+    "delete_tags",
+    help=f"Delete tags from the Database. Any tagged proposals will have the tags removed.",
+)
+@click.argument("tags_to_delete", required=True, nargs=-1)
+def Delete_tags(tags_to_delete):
```
lukegb, line 160: drive-by nit

```suggestion
def delete_tags(tags_to_delete):
```

apps/cfp/tasks.py
```diff
@@ -150,3 +150,31 @@ def create_tags(tags_to_create):
...
+            app.logger.info(
+                f"'{tag_name}' is still applied to the following proposals: {tagged_proposals}"
+            )
+
+        db.session.delete(tag)
+        tags_deleted += 1
+        app.logger.info(f"'{tag_name}' added to session.")
+
+    db.session.commit()
+    app.logger.info(f"Successfully created {tags_deleted} new tags.")
```
lukegb, line 180:
```suggestion
    app.logger.info(f"Successfully deleted {tags_deleted} tags.")
```

apps/cfp/tasks.py
```diff
@@ -150,3 +150,31 @@ def create_tags(tags_to_create):
 
     db.session.commit()
     app.logger.info(f"Successfully created {tags_created} new tags.")
+
+
+@cfp.cli.command(
+    "delete_tags",
+    help=f"Delete tags from the Database. Any tagged proposals will have the tags removed.",
```
marksteward, line 157: I guess "Any tagged proposals will have the tags removed" was a first draft?

apps/cfp/tasks.py
SamLR: not sure what happened there

apps/cfp/tasks.py
SamLR: done

apps/cfp/tasks.py
SamLR: no, the tags have got the correct cascade setting so deleting the tag removes it from the list (at least that's what local testing did)

marksteward: Ahh, I read a `return` that isn't there. Maybe it should say something like f"'{tag_name}' will be removed from the following proposals: {tagged_proposals}".

lukegb (approved):
LGTMBWDIK (looks good to me, but what do I know)

apps/cfp/tasks.py
SamLR: Good point, :172 is now? `f"'{tag_name}' will be removed from the proposals with ids: {tagged_proposals}"`

apps/cfp/tasks.py
```diff
@@ -150,3 +150,31 @@ def create_tags(tags_to_create):
...
+    for tag_name in tags_to_delete:
+        tag = Tag.query.filter_by(tag=tag_name).one_or_none()
+        if not tag:
+            app.logger.info(f"Couldn't find tag: '{tag_name}' exiting.")
+            return
+
+        if tag.proposals:
+            tagged_proposals = [p.id for p in tag.proposals]
+            app.logger.info(
+                f"'{tag_name}' will be removed from the proposals with ids: {tagged_proposals}"
```
jayaddison, line 172: Nitpicky: would it be helpful to mention each proposal `title` as well?  (thinking: could provide a useful sense-check during tag removal)

SamLR: probably not? This will hopefully be a pretty rare manual task (after I do what I need to do which is update the current list of tags to match those selected by the content team) and untagged proposals can be found via the UI. 

In theory deleting a tag will only happen when it's not useful: i.e. has very few (probably no) members OR so many members it's pointless.

jayaddison: Okie doke, just checking.  Seems reasonable!

PR #1311, opened 2024-01-21
Volunteering updates

jellybob:
Update volunteering content:

* Complete rewrite of the volunteering roles page
* Make the top level volunteering page a bit more approachable
  * Possibly controversial is dropping the golden rules - this feels like a thing to communicate during onboarding, rather than the opening salvo on these pages effectively being "you're going to burn out, tell us when you do"

### DIFF
--- a/templates/about/volunteering.md
+++ b/templates/about/volunteering.md
@@ -6,35 +6,25 @@ EMF is a non-profit event, entirely organised by volunteers. Nobody is paid to w
 
 We’d like everyone to consider volunteering to help the event run smoothly, even if it is only for an hour or two.
 
-<small id="footnote">* We provide free tickets for qualified First Aid staff who must work a minimum of three eight-hour shifts during the event.</small>
-
-## How volunteering works
-We understand that everyone who helps with EMF has other commitments and we plan ahead to cope with this.
-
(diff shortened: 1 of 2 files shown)


templates/about/volunteering.md
```diff
@@ -6,39 +6,33 @@ EMF is a non-profit event, entirely organised by volunteers. Nobody is paid to w
 
 We’d like everyone to consider volunteering to help the event run smoothly, even if it is only for an hour or two.
 
-<small id="footnote">* We provide free tickets for qualified First Aid staff who must work a minimum of three eight-hour shifts during the event.</small>
-
-## How volunteering works
-We understand that everyone who helps with EMF has other commitments and we plan ahead to cope with this.
+To be notified when we're looking for help you can sbuscribe to the volunteering list below:
```
lukegb, line 9: subscribe

jellybob: Thank you!

russss:
I'm not a fan of the blue links, sorry. Can you remove that from this PR? I'll see if we can do something else for link visibility.

jellybob:
Done, this is just volunteer copy updates now.

russss (approved):

Issue #1240, opened 2022-06-09
"Your upcoming shifts" page should be sorted by date/time


sde1000:
The "Your upcoming shifts" page currently appears to be randomly (?) ordered; this led to people turning up for bar shifts at the wrong time because they were looking at the first one in the list! It should be sorted by shift start date/time.

ormiret:
I did put in sorting for these lists in 5287c5083ff901c44fcbaee48f1ab14d818799ad (before later killing the whole page as we were still getting reports of it having the wrong times - don't know if they were delayed reports from it not being sorted or if there is more wrong there).

sde1000:
Hmm. Possible timezone stuff? Possible lexical sorting? Would have to look closer.

jayaddison:
Beginning with a series of three volunteer shifts `102`, `103` and `104`, the shift with ID `102` was updated in the `emf_site` DB locally to have it start between `103` and `104`.

The results were that the shifts appeared in the correct order on the signup page ( `/volunteer/schedule` ), with the earliest first by `start` time.

![image](https://github.com/emfcamp/Website/assets/55152140/1696bd69-167a-4529-8630-71db64e734bb)

(in other words: the IDs were re-ordered and did not remain monotonic in the HTML, a successful result here)

I don't _think_ that timezones could/should be a factor here, since this is rendered server-side and uses database-backed timestamps.. even so I'll do a little more digging to build confidence in that.

jayaddison:
Oh, hold on: I tested the wrong page.  Typical.  I'll sign up for some of these shifts in the test environment and then check the "Your upcoming shifts" page...

jayaddison:
From shuffling the data a few times, I haven't been able to replicate sort-order problems on the "Your upcoming shifts" page (after restoring it), so I believe this can be closed.

Interactive SQL session enclosed at the end of this comment to (hopefully) validate the test methodology, which involved shuffling the shift records (I repeated this a few times).

I considered the possibility that user-device timezone settings could be relevant -- perhaps due to organizational editing of events during the fesival, for example - but I don't believe that that can be the case, since the datepickers are timezone-unaware, and the records themselves are also stored without timezone information (that is: the only timeframe of reference, both in the UI and in the database, is event-local).

![image](https://github.com/emfcamp/Website/assets/55152140/d5e13a25-919e-43d8-bb31-7a448d6b2014)

[...]

PR #1312, opened 2024-01-21
testing: update volunteer shift test data for Y2024

jayaddison:
In preparation for QA/testing in #1240.


jayaddison:
Ah: `config_date("EVENT_START")` exists; TODO: use that (minus the time component) instead as the origin date here.

PR #1325, opened 2024-01-23
Add PATCH API to set video URLs on proposals

### LINKED ISSUE #1306: Video API
API access for setting video URLs so that video team can automate linking and uploads

lukegb:
Fixes #1306.

VIDEO_API_KEY needs to be set to something in the config.


lukegb:
I debated with myself about whether we should be able to set the "will be recorded" bit as well but I didn't do that in the end. Maybe we should be allowed to do that?

wlcx:
Ehh, I think we (voc) should be being told that rather than setting it

wlcx:
Whether or not we should be able to set it as individual users is another thing

lukegb:
> Ehh, I think we (voc) should be being told that rather than setting it

This was mostly for late-breaking stuff (I suspect only people _not_ wanting to be recorded where previously they'd said they would) but I think it's easier to just do this some other way and fix it later.

wlcx:
Eh, I’m open. Wonder if anyone else has any strong feelings?

russss:
There's already an [interface](https://github.com/emfcamp/Website/blob/main/apps/schedule/base.py#L297) for the herald or whoever's managing the stage to change this flag (along with other fun things) so you shouldn't need to worry about it.

russss (approved):

wlcx:
@russss 👍

Issue #1214, opened 2022-06-01
Be clearer about ProductGroup.type and Purchase.type


marksteward:
We're using a mixture of checking purchase.product.parent.type and purchase.type (the polymorphic identity). We should only use one.

`ProductGroup.type` also isn't validated, so if someone creates a group with type of e.g. `admission` instead of `admissions`, the purchases created under it won't be AdmissionTickets[[1](https://github.com/emfcamp/Website/blob/94b6c1d/models/basket.py#L205-L210)]. We probably want the hierarchy to be enforced all the way down.

marksteward:
We also show `.price_tier.name` in Purchases's `__repr__`, and not the polymorphic class.

PR #1326, opened 2024-01-23
Provide a product group type dropdown when creating/editing them

### LINKED ISSUE #1322: ProductGroup type should be validated on creation
When adding a ticket to a basket, the type of the enclosing ProductGroup is used to determine whether something is an admission ticket, a parking/caravan ticket, or something else (currently merchandise or rental I believe). That happens here:

https://github.com/emfcamp/Website/blob/39593dae98abc32050e12a3a72b2b3a91ffa03b9/models/basket.py#L203-L209

When creating a ProductGroup, the type is currently just a free-text field, which frequently causes issues when creating these. It should be a dropdown with the following options:

* admissions
* campervan
* parking
* merchandise
* rental

See also #1214

lukegb:
Resolves #1322.

...also fix a bug when viewing payments in the admin panel where it reports "is not None" as an error because "is" has a specific meaning in Jinja2.

### DIFF
--- a/models/basket.py
+++ b/models/basket.py
@@ -9,8 +9,8 @@ from sqlalchemy.orm import joinedload
 from main import db
 from . import Currency
 from .exc import CapacityException
-from .product import PriceTier, Voucher
-from .purchase import Purchase, Ticket, AdmissionTicket
+from .product import PriceTier, Voucher, PRODUCT_GROUP_TYPES_DICT
+from .purchase import Purchase
 
 
 class Line:
@@ -201,12 +201,14 @@ class Basket(MutableMapping):
                     line.tier.issue_instances(issue_count)
 
                     product = line.tier.parent
-                    if product.parent.type == "admissions":
-                        purchase_cls = AdmissionTicket
-                    elif product.parent.type in {"campervan", "parking"}:
-                        purchase_cls = Ticket
-                    else:
-                        purchase_cls = Purchase
+                    product_group_type = PRODUCT_GROUP_TYPES_DICT.get(
+                        product.parent.type
+                    )
+                    purchase_cls = (
+                        product_group_type.purchase_cls
+                        if product_group_type
+                        else Purchase
+                    )
 
                     price = line.tier.get_price(self.currency)
                     purchases = [
--- a/models/product.py
+++ b/models/product.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from dataclasses import dataclass
 from decimal import Decimal
 from collections import defaultdict
 from datetime import datetime, timedelta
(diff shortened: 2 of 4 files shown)


models/basket.py
```diff
@@ -201,12 +201,14 @@ def create_purchases(self):
...
-                    else:
-                        purchase_cls = Purchase
+                    product_group_type = PRODUCT_GROUP_TYPES_DICT.get(
+                        product.parent.type
+                    )
+                    purchase_cls = (
+                        product_group_type.purchase_cls
+                        if product_group_type
+                        else Purchase
+                    )
```
jayaddison, line 211: Could combine these two, by providing a dict lookup fallback/default?

```suggestion
                    purchase_cls = PRODUCT_GROUP_TYPES_DICT.get(
                        product.parent.type,
                        Purchase,
                    )
```

(although arguably less clear, partly due to the text wrapping to fit within line limits)

jayaddison: Ah, nevermind; I see that there's an additional level of indirection for the `purchase_cls` lookup.  Hrm.  I'll ponder that a bit, but this seems OK as-is, and my suggestion's not valid.

lukegb: The problem is that the purchase class is an attribute on the thing in the dict, rather than directly being an element of it, so this doesn't work in and of itself.

templates/admin/payments/payment.html
```diff
@@ -16,7 +16,7 @@ <h2>Payment {{ payment.id }}</h2>
 {% endif %}
 {% if payment.provider == 'banktransfer' and payment.state == 'inprogress' %}
 <a class="btn btn-primary" href="{{ url_for('.reset_expiry', payment_id=payment.id) }}">Reset expiry</a>
-<a class="btn btn-warning {% if payment.reminder_sent_at is not None -%} disabled {%- endif %}" href="{{ url_for('.send_reminder', payment_id=payment.id) }}">Remind</a>
+<a class="btn btn-warning {% if payment.reminder_sent_at is not none -%} disabled {%- endif %}" href="{{ url_for('.send_reminder', payment_id=payment.id) }}">Remind</a>
```
jayaddison, line 19: `is none` was news to me, but yep, that [checks out](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-tests.none) - nice find.

models/product.py
```diff
@@ -66,6 +67,23 @@ def one_or_none(result):
...
 
 
+@dataclass(frozen=True)
+class ProductGroupType:
+    slug: str
+    name: str
+    purchase_cls: type[Purchase]
+
+
+PRODUCT_GROUP_TYPES = [
```
jayaddison, line 77: I'm muddling my way back through the design factors for the change: is there a reason these should go here instead of alongside the [similar-ish constants in `purchase.py`](https://github.com/emfcamp/Website/blob/f0b02a885fe2db4ae3045269dba9f61e6bd6ec8a/models/purchase.py#L12)?

russss: That list doesn't seem to be used at all, so it can probably be removed.

The root issue here is that there's inevitable duplication because we have a polymorphic type hierarchy under Purchase. Which of those Purchase types is created is determined by the Product's `ProductGroup.type`. And the point of #1214 (I haven't checked how relevant it still is) is that we should probably only be checking one of those.

Anyway I'm happy that this PR is a net improvement to the situation.

russss (approved):

Issue #1328, opened 2024-01-24
Wise bank import does not populate payee name


russss:
Payee name was showing as "None" on the bank payment screen - these accounts were auto-populated by the Wise import function:

<img src="https://github.com/emfcamp/Website/assets/45057/0fea7e6a-5ed1-4f8d-a95d-59477ba6b115" width="300">

Additionally, the payment config screen (`/admin/payment-config`) does not show this problem.

We should:

* Import the payee name (presumably this is possible) or require manual entry
* Show the payee name on the payment config screen
* Prevent a bank account from being activated on the payment config screen if it doesn't have a payee name.
* Throw an exception in the payment flow if the payee name or any other required payment details are not present.

No issue for 2024, I've manually populated these in production.

jayaddison:
Ok; we're pulling the incorrect field names here; the `title` field that we attempt to read from is from the new-format [`account-details` API](https://docs.wise.com/api-docs/api-reference/bank-account-details); previously the relevant field was, I think, named `accountHolderName` in the [(now-deprecated) `borderless-accounts` API](https://docs.wise.com/api-docs/api-reference/deprecated).

This would require an update to `pywisetransfer`.

jayaddison:
Closing - I believe this was fixed by #1329.

PR #1329, opened 2024-01-24
[untested] Wise bank account retrieval: update payee source field name

jayaddison:
I _believe_ this should help for #1328 - it's untested, but I'd prefer to open this now while it's fresh in my mind.


jayaddison:
@russss I tried to test this the other day, but the Wise sandbox accounts don't have bank accounts linked to them by default, and I couldn't seem to complete the add-bank-account-link workflows in the sandbox.  I'm relatively confident that this is the correct fix, but could wait for future testing opportunities if we want to be certain.

russss (approved):

PR #1331, opened 2024-01-24
Differentiate live-in vehicle tickets from parking tickets

### LINKED ISSUE #1193: Live-in vehicle tickets are confusingly similar to parking tickets
The live-in vehicle ticket says "Parking ticket" at the top in large bold letters, and "Live-in vehicle ticket" at the bottom in much smaller letters.

This is confusing. It confused me. Live-in vehicle tickets should say "Live-in vehicle ticket" at the top.

We shouldn't change this for 2022 as we've already sent the tickets out and this would result in two different variants of tickets for volunteers to check, but it needs to be fixed for next time

lukegb:
These now use a separate template, but at present the text is identical. The only difference is that one says "Live-in vehicle ticket" and the other says "Parking ticket".

Fixes #1193.

- Screenshot of webpage: https://github.com/emfcamp/Website/assets/246745/b3a31b22-0ef5-45eb-837f-83b7523b3f79
- [As printed to PDF](https://github.com/emfcamp/Website/files/14042308/Electromagnetic.Field.pdf)

### DIFF
--- a/apps/common/receipt.py
+++ b/apps/common/receipt.py
@@ -28,9 +28,8 @@ def render_receipt(user, png=False, pdf=False):
 
     admissions = purchases.filter(ProductGroup.type == "admissions").all()
 
-    vehicle_tickets = purchases.filter(
-        ProductGroup.type.in_(["parking", "campervan"])
-    ).all()
+    parking_tickets = purchases.filter(ProductGroup.type == "parking").all()
+    campervan_tickets = purchases.filter(ProductGroup.type == "campervan").all()
 
     tees = purchases.filter(ProductGroup.type == "tees").all()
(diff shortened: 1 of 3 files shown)


lukegb:
Open questions:

* What should the live-in ticket say?
* Should we differentiate the live-in ticket more so it can be identified easier at a glance/distance (add a pattern to the background of the top of the page, or similar - something visible, that works on B&W lasers and inkjets without wasting too much toner or printer ink)?

wlcx:
> Should we differentiate the live-in ticket more so it can be identified easier at a glance/distance

Perhaps just a big `P` or `L` somewhere? Something like:

image

marksteward:
I've rebased and merged the initial commit for this as d11384f

PR #1345, opened 2024-01-29
Update stripe to latest package/api version

wlcx:
Not suggesting that we necessarily merge this in time for ticket sales!

Issues:
- [x] Check payment/refund with a test stripe account
- [x] Why does mypy thing `stripe_client` is `None|Any`?


wlcx:
Hang on, why not just instantiate the stripe client each time? Are we very concerned with performance? Less global state feels better...

marksteward:
It's more about connecting the config to the library in one place, rather than performance. You can replace it with a `get_stripe_client` function but is that really any better?

wlcx:
Yeah that's what I was thinking. In terms of type checking properly and avoiding globals, I'd have thought so!

russss:
Yes, happy to merge it now, we'll test thoroughly before we need to use it again anyway.

Issue #1254, opened 2022-06-17
may_record doesn't update in admin proposal page


marksteward:
I assume it's the standard checkbox-doesn't-have-a-value-unless-ticked issue. We've switched to using `.raw_data` recently, but this shouldn't need to be so complex.

PR #1346, opened 2024-01-30
More minor CfP changes

lukegb:
- Using raw_data on a checkbox doesn't work, so we can only set (but not unset) may_record. Just use .data and hope that there aren't weird problems if the form field doesn't show up (#1254)
- Update the field names in the CfP finalisation template for the split requirements fields.
- Make the family_friendly field in the finalisation template wider (maybe the checkbox should be left aligned rather than centered?): 
  - before: before
  - after: after


apps/cfp_review/forms.py
```diff
@@ -124,8 +124,7 @@ def update_proposal(self, proposal):
 
         if self.needs_laptop.raw_data:
             proposal.needs_laptop = self.needs_laptop.data
-        if self.may_record.raw_data:
-            proposal.may_record = self.may_record.data
+        proposal.may_record = self.may_record.data
```
marksteward, line 127: Would this also apply to needs_laptop?

apps/cfp/views.py
```diff
@@ -97,7 +98,7 @@ class YouthWorkshopProposalForm(ProposalForm):
     attendees = StringField("Attendees", [DataRequired()])
     cost = StringField("Cost per attendee")
     participant_equipment = StringField("Attendee equipment")
-    age_range = StringField("Age range")
+    age_range = SelectField("Age range", choices=AGE_RANGE_OPTIONS)
```
marksteward, line 101: Maybe remove the 16+ and adult options for this case?

apps/cfp_review/forms.py
lukegb: No; needs_laptop is a SelectField between 0/1, rather than a checkbox.

Maybe this should be a SelectField too?

apps/cfp/views.py
lukegb: Done.

russss:
I think we decided to keep the free-text age input in the CfP for the moment, so can you remove that change? I think the rest is good to merge.

russss (changes requested):
I think we decided to keep the free-text age input in the CfP for the moment, so can you remove that change? I think the rest is good to merge.

lukegb:
Dropped that commit out of the chain.

russss (approved):

PR #1358, opened 2024-02-04
Rename "finished" CfP state to "finalised"

russss:
This also adds an is_accepted columnproperty to Proposal which removes a lot of `Proposal.state in ['accepted', 'finalised']`.

### DIFF
--- a/models/cfp.py
+++ b/models/cfp.py
@@ -363,6 +363,8 @@ class Proposal(BaseModel):
     state = db.Column(db.String, nullable=False, default="new")
     type = db.Column(db.String, nullable=False)  # talk, workshop or installation
 
+    is_accepted = column_property(state.in_(["accepted", "finished"]))
+
     # Core information
     title = db.Column(db.String, nullable=False)
     description = db.Column(db.String, nullable=False)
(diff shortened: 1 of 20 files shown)


models/cfp.py
```diff
@@ -363,6 +363,8 @@ class Proposal(BaseModel):
     state = db.Column(db.String, nullable=False, default="new")
     type = db.Column(db.String, nullable=False)  # talk, workshop or installation
 
+    is_accepted = column_property(state.in_(["accepted", "finalised"]))
```
SamLR, line 366: [nitpick] maybe `is_to_schedule` to make it clearer that this isn't just the `accepted` state?

russss: I think the general meaning of "accepted" outweighs the potential confusion here. Finalised proposals are still accepted proposals.

If we wanted to clarify this further then I would probably advocate renaming the "accepted" state as well, but I can't really think of what to call it ("accepted_unfinalised"? Should "finalised" actually be "accepted_finalised"?) and I'm not sure it's worth it. I think this change is still a net reduction in confusion.

SamLR: that's fair and I think it should still be relatively clear that this isn't straight-up `state == "accepted"`

PR #1359, opened 2024-02-05
Update /sponsor image

joenash:
This PR updates the image displayed at https://emfcamp.org/sponsor to a more recent one. The previous one was the sign from (I believe) 2014, this one is from 2022.

**Questions**:
- I'm removing the original image from the repo, is there any reason to keep the old one?
- This image was sourced from the Hero Photos album, and is by Kari Bellamy. The previous photo had no attribution on the page, should this one?


Jonty:
1) No, that's what git history is for!
2) I don't think so, but we could add it to be nice?

SamLR:
Please add attribution, I believe it's a request of all photos from the heros directory

russss:
We need to design a way to show attribution, then. Which is probably best done in a separate PR.

For what it's worth, all the photos currently shown on the website (and those on archive.org with the `emf_reuse_permission` metadata attribute) have allowed use by us without attribution.

Jonty:
I would just add the attribution as an title tag to the image for now, as we are about to send this to a lot of sponsors imminently - and as russ said, we don't actually need to do this. But we should, because we're nice.

Jonty (approved):

PR #1364, opened 2024-02-09
Rewrite CfP guidance

russss:
This was annoying me. 99% of the guidance about CfP form fields was already available inline - I think I've sorted the remaining 1%.

I then rewrote a better description of the anonymous review process. I think all of the points which were in the FAQ are covered elsewhere. FAQs are bad.

Screenshot 2024-02-09 at 13-21-28 Call for Participation Process


SamLR:
Liedra & I have given it a once over and looks fine. We might want to add a link at the bottom to redirect folks back to the form.

SamLR:
That's fair, either way is good

SamLR (approved):

PR #1366, opened 2024-02-11
Volunteer tweaks

jellybob:
Assorted tweaks to the volunteer system in an attempt to improve usability.

* Rewrote the volunteer schedule Javascript so that it modifies the existing DOM tree passed from the server rather than rebuilding it every time something changes
* Show volunteering info before asking people to sign up so that we can entice them a bit and let them know what it's all about
* Make the info pages markdown
* Make role selection flow better
  * Show the full description for each role rather than forcing people to click more info on each one
  * Redirect to shift selection after choosing roles, hopefully resulting in fewer people not realising they need to sign up for shifts
* If you're not an admin, details buttons are replaced with a simple Sign Up/Cancel button
* If there's a clash between an existing shift and one you're trying to sign up for it will tell you which one


apps/volunteer/main.py
```diff
@@ -19,43 +20,22 @@
 
 
 @volunteer.route("/")
-@feature_flag("VOLUNTEERS_SIGNUP")
 def main():
     if current_user.is_authenticated and Volunteer.get_for_user(current_user):
-        if current_user.shift_entries:
-            return redirect("/volunteer/schedule?signedup")
         return redirect(url_for(".schedule"))
```
jellybob, line 25: This should check for the `VOLUNTEERS_SCHEDULE` flag.

marksteward:
Rebased and merged as 5f125996..eb60596

PR #1369, opened 2024-02-20
add link to current villages list

mscroggs:


wlcx:
I think this will 404 in non-event years thanks to https://github.com/emfcamp/Website/blob/10c849f7c09f104968b039a9a56a55f70dac4fe1/apps/villages/views.py#L64 Which might not be the best experience.

Not quite sure how best to improve that though. We could guard showing this with the same check (`if year == event_year()`) but perhaps it would be best to have that split out somewhere (`if village_registration_available()`?)

mscroggs:
I've added /2024 to the link so that it won't like to a 404 page. Maybe this is ok for now so that we can get this link up and people can start looking at the list of villages?

wlcx:
LGTM

marksteward:
Rebased and merged (44fc74b)

PR #1372, opened 2024-02-21
Add non-admin CFP user view

### LINKED ISSUE #1371: Create a non-admin view of users
when a cfp_admin clicks on a user in a proposal they should get sent to some sort of useful view.

This should show all that user's:
* proposals
  * + message status (& note counts) for their proposals
* email address
* cfp_invite_reason
* Whether they have a ticket
* arrival time if given?

If the a user_id that doesn't have any proposals is set the page should 404

wlcx:
Looks like this:
image

Closes #1371


marksteward:
Rebased and merged (1cb78689)

PR #1373, opened 2024-02-21
Limit number of tickets by amount remaining on voucher, if using

### LINKED ISSUE #1334: Voucher ticket maximum does not affect quantity dropdown on ticket choose page
If you have a voucher which is good for 2 adult tickets, you still have the ability to select up to 10 tickets on the choose page. When you click the next button, you get an error, so this is just a UI issue.

Two people have already pointed this out to me this year so it would be nice to have this fixed, but not a major issue.

wlcx:
Not 100% on this logic, so would appreciate a lookover.

Does it matter if a voucher is used when tickets are still generally available? I think this would mean that the user sees the reduced amount, when in reality they could technically buy more. But that feels like a very rare situation....

Closes #1334

### DIFF
--- a/apps/tickets/choose.py
+++ b/apps/tickets/choose.py
@@ -83,8 +83,11 @@ def main(flow="main"):
         # Empty form - populate products with any amounts already in basket
         form.populate(basket)
 
+    voucher = None
+    if code := session.get("ticket_voucher"):
+        voucher = Voucher.get_by_code(code)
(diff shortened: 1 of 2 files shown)


marksteward (approved):

marksteward:
Yeah, the voucher bypasses the sales state so we should probably limit to the voucher value rather than risk opening up a portal to unlimited tickets.

PR #1375, opened 2024-02-22
Add CFP proposals export

### LINKED ISSUE #1197: Add ability to export data for current CfP review view in CSV / JSON / etc format
Use case is e.g exporting lists of installations for creation of spreadsheets, running lists of things through scripts to generate print materials etc etc

wlcx:
A little messy but does the job.

@prehensile - is this set of fields sufficient or do any need to be included/excluded?

Resolves #1197


marksteward:
Rebased and merged (de325ea)

PR #1380, opened 2024-02-28
User select boxes for the diversity form & add views to track speaker & reviewer diversity

SamLR:
We want to track speaker & reviewer diversity in both cases to track the efficacy of outreach but also in the latter to check that our reviewers' interests cover the topics of our talks & workshops.

To do this we need to make the values requested more uniform so we can more easily track them (currently there are respectively 64 & 96 unique genders & ethnicities from the ~500 submissions).

This reduces the genders to:
* female
* male
* non-binary
* transman
* transwoman
* other

And the ethnicities to (from https://www.ethnicity-facts-figures.service.gov.uk/style-guide/ethnic-groups)
* asian
* black
* mixed
* white
* other

based on some (quick) testing the `guess_gender` and `guess_ethnicity` functions match the bulk of responses we've received (456 out of 502 and 347 out of 428 for them respectively).

I think this is about good to go but we should probably all give it a look on the test server first

Screenshot 2024-02-28 at 22 19 25
Screenshot 2024-02-28 at 22 19 18


russss:
I think gender should just be female/male/non-binary/other. Trans people are the gender they identify as, and providing separate transman/transwoman options will likely not accurately capture that. If we wanted to do that, there should be a separate "do you identify as trangender" checkbox, but I'm not sure we need to.

Otherwise looks good. I'd squash these commits as they'll be rebased onto the main branch.

SamLR:
> I think gender should just be female/male/non-binary/other. Trans people are the gender they identify as, and providing separate transman/transwoman options will likely not accurately capture that. If we wanted to do that, there should be a separate "do you identify as trangender" checkbox, but I'm not sure we need to.

Happy to go with that if we'd prefer. Completely agree on trans folk being the gender they identify as, I'd only added those options because people had been using variants of them in the free text version

wlcx:
> If we wanted to do that, there should be a separate "do you identify as trangender" checkbox, but I'm not sure we need to.

Yeah, I think a "Is your gender identity the same as the sex you were assigned at birth" question would be the way to do this (With yes/no/I prefer not to answer/other) options[1]. I think it might be worth us still collecting this if folks want to share that information.


[1]: https://ruth-dm.co.uk/how-to-ask-about-gender-in-forms-respectfully

marksteward:
Merged with manual rebase

PR #1388, opened 2024-03-09
Encourage villages to make wiki pages

### LINKED ISSUE #1370: Add info to about/villages about wiki pages
If we're not having a wiki, it would be nice to have a page on the website for each village that includes who to contact if you're interested in joining that village (and any other info that we used to recommend putting on the wiki).

If we decide to have a wiki, we can close this.

mscroggs:
Resolves #1370

### DIFF
--- a/templates/_footer.html
+++ b/templates/_footer.html
@@ -1,4 +1,4 @@
-w<footer role="contentinfo">
+<footer role="contentinfo">
   <div id="bottom-menu-container">
     <ul class="menu" role="menu">
     {#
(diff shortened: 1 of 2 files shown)


templates/_footer.html
```diff
@@ -1,4 +1,4 @@
-<footer role="contentinfo">
+w<footer role="contentinfo">
```
russss, line 1: Looks like a "w" has crept in here

russss (changes requested):

templates/_footer.html
mscroggs, line 1:
```suggestion
<footer role="contentinfo">
```

russss (approved):

PR #1391, opened 2024-03-11
Make the volunteer schedule more usable on mobile

jellybob:
* Use `responsive-table` to turn each row into a block element that can be easily read on a mobile device.
* Switch the day tabs for a select box on mobile so they don't take up half the screen.


lukegb (approved):

marksteward:
Rebased as 58e5eef and merged

Issue #1384, opened 2024-03-06
Make separate 'tags' concertina section in `/proposal/<id>`


SamLR:
Add a new section between 'basics' and 'requirements' for 'tags'. This should only contain the tags multi-select but the section title should include any assigned tags for that proposal e.g.:

* `tag ()` -- if no tags set
* `tag (trains)` if one tag set
* `tag (robots, psychology, games)` for multiple

In theory this may produce a list long enough to break the rendering (especially on small screens) so in that case they may need `overflow: text` or similar setting.

If possible sorting the select list to float the chosen tags to the top would be amazing but that's probably a lot to ask...

PR #1397, opened 2024-03-15
improve the tag ui for proposals

SamLR:
move the tags into their own collapse panel & put the selected tags in the header

originally this was going to use checkboxes but that requires a bigger change and if this works for folks I'd rather work on other things :p

default view (note list of tags in header):
Screenshot 2024-03-15 at 20 42 36

open view (still using multiselect field)
Screenshot 2024-03-15 at 20 32 10


wlcx:
Not tested but looks fine to me

wlcx:
Closes #1384 right?

SamLR:
> Closes #1384 right?

it does

jellybob (approved):

PR #1399, opened 2024-03-15
cfp/schedule: Add video recording lost flag

### LINKED ISSUE #1258: "Video recording lost" flag on CfP items
If we have this, we can have a "this video is still being processed" or "sorry, we've got no video for this talk" message on the talk page.

This flag should also be exported to the static export.

wlcx:
Looks like this:
image

Closes #1258


marksteward:
Rebased and merged as 5afdd90

PR #1413, opened 2024-03-30
Better cfp bot messages

SamLR:
Add some emoji to the cfp irc bots so stuff stands out a bit more


apps/cfp/views.py
```diff
@@ -827,7 +827,7 @@ def proposal_messages(proposal_id):
             if channel := app.config.get("CONTENT_IRC_CHANNEL"):
                 # WARNING: don't send personal information via this (the channel is public)
                 irc_send(
-                    f"{channel} New CfP message for {proposal.human_type}: {external_url('cfp_review.message_proposer', proposal_id=proposal_id)}"
+                    f"✉️ {channel} New CfP message for {proposal.human_type}: {external_url('cfp_review.message_proposer', proposal_id=proposal_id)} ✉️"
```
ormiret, line 830: the first ✉️ should be the other side of {channel} or it'll be changing where it tries to send rather than added to the message

marksteward:
There should also be a fake_irccat app that is entirely to provide somewhere for IRCCat messages to go in dev. If that's not working you might have it disabled in your docker-compose.

marksteward:
(Didn't send first time.) DEBUG isn't a feature_flag, it's a config item (or attribute on Flask).

apps/cfp/views.py
SamLR: Good catch, I've made `channel` an explicit argument in `irc_send` to make this clearer

SamLR:
@marksteward I forgot that container was there, I've removed the `debug` statement :)

PR #1418, opened 2024-04-01
Push notifications

jellybob:
Work in progress PR, this handles the absolute basics of registering for push notifications, and being able to send a notification to an endpoint.

Things still to do:

* [ ] Provide a reasonable UI
* [ ] Test on multiple platforms (so far only Safari on macOS has been tested)
* [ ] Better handling of service worker updates
* [ ] Being able to open a specific page, and customisable title/body/actions


wlcx:
Haven't looked through super closely, but looks great and could also fit really well alongside APNS push notifications for apple wallet tickets...

jellybob:
This is probably in a reasonable state for testing now. No notifications are being queued other than those requested with the test button, but I think there's some value in wider testing of that basic functionality given how finicky WebPush has been for me so far.

You'll need to set `WEBPUSH_PUBLIC_KEY` and `WEBPUSH_PRIVATE_KEY`. You can generate a key pair with `npx web-push generate-vapid-keys`.

apps/notifications/jobs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    for job in jobs:
+        deliver_notification(job)
+        db.session.add(job)
+
+    db.session.commit()
+
+
+@scheduled_task(minutes=15)
+def queue_content_notifications(time=None) -> None:
```
jellybob, line 54: I deeply dislike the approach being taken here, if anyone with sufficient SQLAlchemy knowledge wants to turn it into a single query with some joins I'd be very happy.

russss:
Closing as stale, sorry.

PR #1419, opened 2024-04-03
Fixed broken link: /volunteering/ -> /about/volunteering (on Travel page)

leowilkin:
I fixed a broken link found on the [travel page](https://www.emfcamp.org/about/travel) that incorrectly linked [Volunteering](https://www.emfcamp.org/about/volunteering)

**Before:**
`You won't be allowed on site outside those times unless you've [volunteered](https://www.emfcamp.org/volunteering) to help with setup or teardown.`

**After:**
`You won't be allowed on site outside those times unless you've [volunteered](https://www.emfcamp.org/about/volunteering) to help with setup or teardown.`


jellybob (approved):
Thanks!

Issue #1406, opened 2024-03-25
Configurable VAT rate for products


russss:
We're going to be selling a lower self-service "sponsorship" tier through the website. As this does not entitle the sponsor to anything other than a mention on the website, it is outside the scope of VAT.

* There should be a VAT rate column on the product. This should default to a value which is set in the configuration (currently 20%).
* This column can be nullable where null means "VAT Exempt"
* The receipt/VAT invoice page needs to identify this and show the correct VAT amounts

We do ideally also need some report for this but that doesn't need to be part of this issue.

Mattsi-Jansky:
I can give this a crack. Just to rephrase the requirements to make sure I understood correctly:

* Add a configurable VAT rate column
* On the invoice page, when the VAT rate column is null disable VAT altogether or otherwise set it to zero

I wonder if the new column could be a boolean instead, if the VAT rate is not likely to change per product? Something like `is_subject_to_vat`

russss:
It turns out we do also need to handle items which are zero-rated for VAT (namely, children's T-shirts) and this is different from exempt, so I think it's probably best if we do have an actual rate column.

It might be best if we just use an enum ("Standard Rate", "Zero Rate", "Exempt") though, rather than putting the VAT rate in the column.

marksteward:
I'd rather store the percent, in case we cross over with a VAT change in the future.

russss:
I'm trying not to think about that possibility but if we do want to consider that then we probably need to store the VAT percentage on the price tier, not on the product. It gets complicated because there are potentially three zero-VAT states which we need to consider - zero-rated, exempt, and outside the scope. (Although I think exempt and outside the scope are functionally identical).

russss:
Yup, I got it the wrong way round. Zero rated and exempt are functionally identical from our perspective. And donations are outside the scope, not exempt.

So we could do this with a nullable column where null means outside the scope, and I think it would be best to put it on the PriceTier.

Mattsi-Jansky:
Made a start. Facing some issues with setting the default value from a configurable because the application configuration isn't available from the `models` scope, if I understood the following error right: 

```
models/product.py:273: in Product
    db.Float, nullable=True, default=app.config["DEFAULT_VAT_RATE"]
/root/.cache/pypoetry/virtualenvs/website-9TtSrW0h-py3.11/lib/python3.11/site-packages/werkzeug/local.py:311: in __get__
    obj = instance._get_current_object()
[...]

PR #1434, opened 2024-04-15
Add VAT Rate field, enabling VAT exemption for specific products/price tiers

Mattsi-Jansky:
Addresses #1406

Few notes:

- Because the database doesn't seem to support storing decimals I store the VAT rate as a string and convert it at runtime.
- VAT rate can be modified from the admin panel
- The decimal arithmetic can result in some odd figures - e.g. £10 excluding a VAT of 20% gives you 8.3333 recurring, yet a value excluding vat of £8.33 with VAT included comes to 9.9999 recurring. I made the decision to just return raw decimals from these calculations and let the view decide how to round them.


marksteward:
Thanks for this! Rebased with some extra fixes as 84c6b9a and merged.

PR #1436, opened 2024-04-16
poetry: update gunicorn to v21.2.0

jayaddison:
Based on the `gunicorn` [security support policy](https://github.com/benoitc/gunicorn/security), I think we should update to the most-recent version [v21.2.0](https://github.com/benoitc/gunicorn/commit/ab9c8301cb9ae573ba597154ddeea16f0326fc15) - it has been the latest version since its release on 2023-07-19, and seems relatively stable based on searching [issues that mention it](https://github.com/benoitc/gunicorn/issues?q=is%3Aissue+is%3Aopen+21.2.0).

There is one bugreport https://github.com/benoitc/gunicorn/issues/3038 - a possible regression since the v21.0.0 release series that isn't yet resolved.

Additionally there have been a few reports ([ref](https://github.com/benoitc/gunicorn/issues/3046), [ref](https://github.com/benoitc/gunicorn/issues/3120)) of problems with `eventlet` (aka `geventlet`) - we don't use that worker-type though; [we use `gthread`](https://github.com/emfcamp/Website/blob/1bf11f19de6bdcad7bb26e8a287c076d9dcc7565/docker/prod_entrypoint.sh#L19).

NB: The bugreport mentioned above _does_ appear to be related to a `gthread` regression.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1359,17 +1359,17 @@ test = ["objgraph", "psutil"]
 
 [[package]]
 name = "gunicorn"
-version = "20.1.0"
+version = "21.2.0"
 description = "WSGI HTTP Server for UNIX"
 optional = false
 python-versions = ">=3.5"
 files = [
-    {file = "gunicorn-20.1.0-py3-none-any.whl", hash = "sha256:9dcc4547dbb1cb284accfb15ab5667a0e5d1881cc443e0677b4882a4067a807e"},
-    {file = "gunicorn-20.1.0.tar.gz", hash = "sha256:e0a968b5ba15f8a328fdfd7ab1fcb5af4470c28aaf7e55df02a99bc13138e6e8"},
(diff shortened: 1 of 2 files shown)


jayaddison:
> There is one bugreport [benoitc/gunicorn#3038](https://github.com/benoitc/gunicorn/issues/3038) - a possible regression since the v21.0.0 release series that isn't yet resolved.

Update: I don't think we're affected by this because we don't configure `--max-requests` when running `gunicorn`.

russss (approved):

russss:
I'll merge anyway to avoid any security warnings, thanks!

jayaddison:
Thanks - note that a fix for CVE-2024-1135 has been merged into `gunicorn`, but is pending a release (so dependabot alerts about that particular CVE won't be solved by this).

PR #1441, opened 2024-04-21
Better self refund

### LINKED ISSUE #900: Allow refund system to be used for normal refunds
The refund system is a bit opaque at the moment and we should make this clearer to the ticket purchaser. If they request a refund for a payment, they should be able to select which specific tickets they'd like refunded. We should show the total amount of the refund.

Users should not be able to request a refund for tickets they have transferred to other people (we should ask them to transfer those tickets back before requesting a refund). This requirement should not apply in the case that the event is cancelled - in this case (only) the original purchaser should be able to request a refund for the full amount of their payment.

Users should not be able to request a refund for checked-in tickets. We can do this through the backend in exceptional circumstances.

I'm doing some initial work for this while sorting out 2020 refunds but I'm only concerned about the (unusual) cancellation refund flow in this case.

SamLR:
Closes #900 

This is primarily the functionality. The CSS still needs improvement and the copy is lacking (e.g. at the moment the 'illness' state merely adds an extra line to the /refund page copy)

This adds a new state machine which doesn't currently get initialised properly all the time for some reason.

The previous 'request refund' flow still works in 'cancellation' mode

Screenshot 2024-04-21 at 01 48 53

The user refund form:

Screenshot 2024-04-21 at 17 24 29

The previous (now 'cancellation') form

Screenshot 2024-04-21 at 01 49 48


apps/admin/payments.py
```diff
@@ -390,41 +384,17 @@ def refund(payment_id):
...
     if form.validate_on_submit():
         if form.refund.data or form.stripe_refund.data:
 
             payment.lock()
 
             purchases = [
-                f._purchase for f in form.purchases if f.refund.data and not f._disabled
+                purchases_dict[f.purchase_id.data]
+                for f in form.purchases
+                if f.refund.data and purchases_dict[f.purchase_id.data].is_refundable
```
jayaddison, line 395: Should this be a method call?

```suggestion
                if f.refund.data and purchases_dict[f.purchase_id.data].is_refundable()
```

models/payment.py
```diff
@@ -147,6 +149,15 @@ def get_export_data(cls):
 
         return data
 
+    def is_refundable(self, override_refund_state_machine=False) -> bool:
```
jayaddison, line 152: I'd suggest inverting the argument:

```suggestion
    def is_refundable(self, by_admin=False) -> bool:
```

The verbose reason: payment processing software often uses [state machine logic](https://en.wikipedia.org/wiki/Finite-state_machine) to co-ordinate the lifecycle of payments, purchases, refunds, disputes, etc; it'd be nice to avoid a naming conflict with that.

templates/account/purchases/payments.html
```diff
@@ -30,7 +30,7 @@
           <a href="{{ url_for('payments.stripe_cancel', payment_id=p.id) }}" class="btn btn-warning">Cancel</a>
           {% endif -%}
 
-          {% if feature_enabled('REFUND_REQUESTS') and p.state == "paid" %}
+          {% if REFUND_STATE != "off" and p.state in ["paid", "partrefunded"] %}
```
jayaddison, line 33: Checking the change here: previously it was only the case that admins could partially refund a purchase -- but now users will be able to do that, and that means that `partrefunded` purchases should be included (for further refund requests?) here.

templates/payments/event-cancelled-refund-request.html
```diff
@@ -0,0 +1,60 @@
...
+   please <a href="mailto:{{ config['TICKETS_EMAIL'][1] }}?subject=Refund Query (WEB-{{payment.id}})">drop us an email</a>.</p>
+
+<div class="well">
+<form method="post" action="{{ url_for('.payment_refund_request', payment_id=payment.id, currency=currency) }}">
+{{ form.hidden_tag() }}
+  <fieldset>
+    <legend>Donation</legend>
+    <p>If you want, you can donate some of your payment to help us recoup our losses for this year
+        and fund our next event. This is completely optional &mdash; we have enough money in the bank
+        to survive until 2022.</p>
```
jayaddison, line 22:
```suggestion
        to survive until 2026.</p>
```

templates/payments/refund-request.html
```diff
@@ -12,18 +12,31 @@ <h2>Refund your payment</h2>
...
-    <p>If you want, you can donate some of your payment to help us recoup our losses for this year
-        and fund our next event. This is completely optional &mdash; we have enough money in the bank
-        to survive until 2022.</p>
-    {{ render_field(form.donation_amount, min=0, max=payment.amount, value=0,
-          donation_range_payment=payment, donation_range_currency=CURRENCY_SYMBOLS[currency]) 
-    }}
-  </fieldset>
+
+
+  <p>Please select any and all purchases you would like to receive a refund for.</p>
```
jayaddison, line 27: Not sure if this is necessary or just nitpicking, but I'd suggest:

```suggestion
  <p>Please select any and all purchases you would like to request a refund for.</p>
```

Also not 100% sure whether it should be plural (`a refund for` / `refunds for`).  I'll try to consult a language reference after some other code review.

jayaddison: I suppose that in the case of multiple selected items, a single refund is issued, so in that sense singular (`a refund`) is correct.

templates/payments/event-cancelled-refund-request.html
jayaddison: (maybe, unless this template is now entirely for historical purposes; in which case perhaps we should rely on `git` history for that instead, and drop it?)

models/purchase.py
```diff
@@ -56,6 +57,7 @@ class Purchase(BaseModel):
     # Financial FKs
     payment_id = db.Column(db.Integer, db.ForeignKey("payment.id"))
     refund_id = db.Column(db.Integer, db.ForeignKey("refund.id"))
+    refund_request_id = db.Column(db.Integer, db.ForeignKey("refund_request.id"))
```
jayaddison, line 60: I've probably just overlooked the relevant flask form / sqlalchemy logic, but could I check where this columns is used/reference?

migrations/versions/7a6816e652f2_add_purchases_to_refund_requests.py
```diff
@@ -0,0 +1,30 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('purchase', sa.Column('refund_request_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(op.f('fk_purchase_refund_request_id_refund_request'), 'purchase', 'refund_request', ['refund_request_id'], ['id'])
+    op.add_column('purchase_version', sa.Column('refund_request_id', sa.Integer(), autoincrement=False, nullable=True))
```
jayaddison, line 21: Sorta curious whether this is planned to have `autoincrement=True` in future? (basically I'm wondering whether the argument needs to be included)

docs/testing.md
```diff
@@ -19,6 +19,31 @@ This is the easiest method to "pay" for a ticket when you're developing. The det
...
+4. install stripe cli `brew install stripe/stripe-cli/stripe`
+5. `stripe login`
+
+#### If the stripe api version is not the latest
+
+1. Get an [ngrok.com](https://ngrok.com) account
+2. install the ngrok client (`brew install ngrok/ngrok/ngrok`)
+3. `ngrok ngrok config add-authtoken <some token>` (the ngrok dashboard should show the token)
+4. `ngrok http http://localhost:2342` this should show the ngrok forwarding domain for you (something like https://d5a7-329-22-333-28.ngrok-free.app)
+5. `stripe webhook_endpoints create --url https://d5f7-149-22-218-28.ngrok-free.app/stripe-webhook --api-version 2020-08-27 --enabled-events '*'`
```
jayaddison, line 37:
```suggestion
4. `ngrok http http://localhost:2342` this should show the ngrok forwarding domain for you (something like https://ffff-111-22-333-44.ngrok-free.invalid)
5. `stripe webhook_endpoints create --url https://ffff-111-22-333-44.ngrok-free.invalid/stripe-webhook --api-version 2020-08-27 --enabled-events '*'`
```

apps/payments/main.py
```diff
@@ -138,6 +182,18 @@ def payment_refund_request(payment_id, currency=None):
...
+        return render_template(
+            "payments/event-cancelled-refund-request.html",
+            payment=payment,
+            form=form,
+            currency=currency,
+            bank_validation_failed=bank_validation_failed,
+        )
+
     return render_template(
         "payments/refund-request.html",
```
jayaddison, line 198: It'd be nice to retain a single template for these two if that's possible.  Understood if it's becoming too difficult to manage the variations using conditions, though.

models/payment.py
SamLR: I would rather not because `by_admin` doesn't really tell you anything about what it's being done whereas this tries to make it much clearer what the actual change is. 

there is definitely a risk of confusion with the other statemachines we have (as well as payment ones) but that's part of the reason for calling out the use of 'refund'

templates/account/purchases/payments.html
SamLR: yes, the intention is that users will be able to refund, e.g. 3 out of 5 tickets

models/purchase.py
SamLR: It's not actually used (at the moment) but it allows us to link back to the request as well. Given that (in theory) when processing a request an admin could select different tickets to refund this means we can refer back the original request and confirm what was requested

migrations/versions/7a6816e652f2_add_purchases_to_refund_requests.py
SamLR: honestly not sure, this is the alembic/versioning code which I tend to leave alone unless I have reason to change

apps/payments/main.py
SamLR: I may revisit this but I'd like to leave it as is for the time being

models/payment.py
jayaddison: Ok, yep.  Any chance of `ignore_event_refunds_state` or something similar instead?

migrations/versions/7a6816e652f2_add_purchases_to_refund_requests.py
jayaddison: Ah, got it.  Yep, that's fine then - thanks.

templates/account/purchases/payments.html
jayaddison: Ok, that part I understood; but specifically here, the idea seems to be that users can request another refund from an already-partly-refunded purchase?

1. User purchases 5 tickets; status = `paid`; this template displays the refund request form (OK).
1. User selects refunds for 3 of 5 tickets; status = `partrefunded`; this template continues to display the refund request form (is that OK?  I think so, probably.  but this is the aspect I'm wondering about).

apps/admin/payments.py
jayaddison: Thanks - there are a few more of these though, all in `main.py` afaict.

models/purchase.py
```diff
@@ -111,6 +113,17 @@ def __repr__(self):
     def is_transferable(self):
         return self.product.get_attribute("is_transferable")
 
+    def is_refundable(self, override_refund_state_machine=False) -> bool:
+        return (
+            (self.is_paid_for is True)
+            and not self.is_transferred
+            and self.payment.is_refundable(override_refund_state_machine)
+        )
```
jayaddison, line 121:
```suggestion
        return (
            self.is_paid_for
            and not self.is_transferred
            and self.payment.is_refundable(override_refund_state_machine)
        )
```

(seems consistent with elsewhere in the codebase)

SamLR: this change is purely there so that mypy is happy (otherwise it sees `is_paid_for` as a column type rather than a boolean

templates/account/purchases/payments.html
SamLR: I don't see why it would be a problem? Probably a question for @russss

models/payment.py
jayaddison: Thanks!

models/purchase.py
jayaddison: Gotcha; frustrating, but fair enough (I had a quick look at adjusting the definition / `column_property` to resolve that, and couldn't immediately find a way to do that; at least not without [making it a Python `@property`](https://docs.sqlalchemy.org/en/20/orm/mapped_sql_expr.html#using-a-plain-descriptor), and apparently that'd incur performance degradation).

jayaddison (approved):
Looks good to me; the limitations of my review are:

  * I don't yet grok the data model in enough detail to reassure myself that the refund-of-a-partial-refund case is OK (but it does appear to be).
  * I haven't tested the changeset locally (but will try to post-review anyway, partly to help address that other limitation).

Jonty:
I can't easily test this due to trainternet but it looks reasonable from a scan - Are we keeping the draggy-refund slider to allow for a donation to be made?

SamLR:
draggy refund slider is only visible if the refunds are being handed out because of cancellation

templates/account/purchases/payments.html
marksteward: It's fine, it was more annoying in the past when refunds required more manual work.

marksteward (approved):

PR #1458, opened 2024-04-27
Add jobs board feature

Mattsi-Jansky:
As discussed elsewhere, the idea here is to improve our sponsorship offerings by creating a job board that we can sell entries for.

image

- It reads from a JSON file and lists the entries from the file. No database changes.
- Clicking any entry takes you to `url` from the JSON object, in a new tab
- Includes feature flag `JOBS_BOARD`. When disabled the page 404s.
- Adds a nav bar link ("Jobs", not sure if there was a better option for the label?) that is also feature flagged


russss:
We decided not to go ahead with this.

PR #1480, opened 2024-05-04
Suggested wording update

katiesteckles:


templates/emails/cfp-slot-scheduled.txt
```diff
@@ -21,7 +26,7 @@ Things to note:
 * You can also update your {{ proposal.human_type }} title and description at any time using the link above, so make sure it sounds great!
 
 {% if proposal.type == 'talk' %}
-* We would strongly prefer that you do not take on-stage Q&A after your talk unless it is, for instance, a panel that requires audience participation - these are hard for us to manage, and frequently unproductive. Please consider taking questions outside the tent, or in the bar instead.
+* We would strongly prefer that you do not take on-stage Q&A after your talk - these are hard for us to manage, and frequently unproductive. We have a dedicated Q&A tent - located near the lounge - where you can do this instead, and direct people to join you there after your talk.
```
jayaddison, line 29: Perhaps we could hyperlink to the Q&A tent location on the map for convenience here?  (note: the link below may not genuinely be where the tent is; I couldn't find it labeled)

```suggestion
* We would strongly prefer that you do not take on-stage Q&A after your talk - these are hard for us to manage, and frequently unproductive. We have a dedicated Q&A tent - [located near the lounge](https://map.emfcamp.org/#16.9/52.04083/-2.376608/m=52.041464,-2.376958) - where you can do this instead, and direct people to join you there after your talk.
```

jayaddison: Ah, drats - this is a `.txt` file, so that wouldn't work.

jayaddison (approved):

templates/emails/cfp-slot-scheduled.txt
crablab, line 29:
```suggestion
* We do not have onstage Q&A at EMF- these are hard for us to manage and frequently unproductive. 
If you'd like to continue the discussion after your talk, the Herald will invite attendees to meet you at the dedicated Q&A tent - located near the lounge - or an alternative venue such as your village or the bar. 
```

TBD location

marksteward:
Rebased and merged as 3878efb3

PR #1487, opened 2024-05-06
2024 volunteer roles

jellybob:
* YAMLise volunteer venues and roles to make editing less painful
* Add/update roles for 2024
* Some light tweaks to the role selection flow


apps/volunteer/main.py
```diff
@@ -16,17 +16,21 @@
...
 def main():
-    if (
-        feature_enabled("VOLUNTEERS_SCHEDULE")
-        and current_user.is_authenticated
-        and Volunteer.get_for_user(current_user)
-    ):
-        return redirect(url_for(".schedule"))
+    if current_user.is_authenticated and Volunteer.get_for_user(current_user):
+        volunteer = Volunteer.get_for_user(current_user)
+        if volunteer is None:
```
ormiret, line 26: I don't think `volunteer` can be `None` here (unless the volunteer is deleted in the DB between the two queries) or the check on line 24 would have failed. 

Could assign `volunteer` on line 24 with walrus: `if current_user.is_authenticated and (volunteer := Volunteer.get_for_user(current_user)):`

PR #1519, opened 2024-05-09
Update /badge redirect

MatStace:
The /badge redirect currently goes to https://wiki-archive.emfcamp.org/2012/wiki/TiLDA which is 404, I figured that it would be better off redirecting to the tildagon site.


jellybob:
Applied directly rather than merging 1 line of code and 3 commits of subsequent merges.

PR #1542, opened 2024-05-15
Vehicle vol role

ormiret:


jellybob:
Merged as part of #1552

PR #1551, opened 2024-05-16
Add mention of bonus meals for KA roles

paullessing:
I believe this also needs to be merged into the DB, once it's approved.


lukegb (approved):

jellybob:
Merged as part of #1552

PR #1554, opened 2024-05-17
Make collecting things more generic

### LINKED ISSUE #434: Checkin improvements
The "checked in at" calculation is wrong, and I don't like all the explicit csrf_token() calls and checks for badge. We also need to extend it to work for t-shirts.

### LINKED ISSUE #1485: Badge/Merch check-in
This year badges and t-shirts are purchased through the ticket store and will need to be checked in the same way that admission tickets are. We'll need to scan the barcode and mark it as issued in the ticket system so it can't be claimed twice.

lukegb:
* Moves `AdmissionTicket.badged_up` and `.checked_in` to `Purchase.redeemed`
  * This is guarded behind a `is_redeemable` attribute; on my dev instance I set this to true on the `admissions` and `merchandise` product groups
    * There's now an admin UI for seeing this field on product groups, as well as on products
* Creates Arrivals Views, which are just a slightly rebranded version of Product Views but for the Arrivals UI
  * Arrivals Views are associated with a permission, if you have the permission you can see that view
  * This is more of a butterfingers check than anything else, all the views show ~all purchases (but not in a consumable/redeemable format)
* Renamed the `arrivals:badge` permission to `arrivals:merch` because I'm touching it (it was that or create `arrivals:tees`, but I don't think that granularity is strictly necessary)
* Some small fixes
  * make fix-syntax now works
  * `flask tickets create_merch` now creates a merch hierarchy, the same as I used for testing. Not really useful in prod.
  * Some cleanup of `flask tickets create` so it doesn't set the `has_badge` attribute, but does now set `is_redeemable`
  * `flask tickets email` now runs inside a test request context, because static_url_for relies on flask-static-digest which now tries to access `request.blueprint` which explodes if it isn't in a request context. It doesn't really seem to matter, though...

Some warnings:

* If something isn't consumable, you can still add it to an ArrivalsView, it'll just break when someone tries to consume it in the arrivals UI. This is probably worth fixing.
* The CSS attributes that we apply during user search on /arrivals don't currently actually do anything to the table, since the styles are overridden by datatable.

Fixes #434
Fixes #1485


marksteward (approved):

marksteward:
Amazing 🥳

marksteward:
(Rebased and merged with cf5873b)

PR #1559, opened 2024-05-17
Swap La Pizzeria's non-existent(?) website for their Instagram

lukegb:
Please check that this is actually the right food vendor, but it looks plausible.


marksteward:
Looking at the docs on the drive, I don't think it is.

marksteward (approved):

PR #1565, opened 2024-05-19
Add CfP scheduling sense-check

### LINKED ISSUE #1564: Sus proposal view
There are currently sometimes valid reasons for us having talks/etc in places/times that we would not automatically schedule them, and sometimes reasons why certain things are omitted. Ideally we fix all that for 2026, but for this year we could do with something that flags these things up so we can't miss them.

I'm thinking a single page that lists all proposals in accepted/finalised state with one of the following true (and it says which):

* Proposal is before Wednesday 12pm, or after Sunday 2am.
* Proposal is after 2am or before 9am.
* Proposal does not have a duration.
* Proposal is in a venue that is not in the allowed list for that type of content.
* Proposal is accepted/finalised and does not have a proposed or scheduled time.
* Proposal has allowed time periods that violate the same rules above.
* What else?

lukegb:
Fixes #1564.

### DIFF
--- a/models/cfp.py
+++ b/models/cfp.py
@@ -770,7 +770,7 @@ class Proposal(BaseModel):
 
     @property
     def potential_end_date(self) -> Optional[datetime]:
-        start = self.start_date
+        start = self.potential_start_date
         duration = self.scheduled_duration
         if start and duration:
             return start + timedelta(minutes=int(duration))
--- a/apps/cfp_review/__init__.py
+++ b/apps/cfp_review/__init__.py
@@ -134,3 +134,4 @@ def cfp_review_variables():
 from . import base  # noqa: F401
 from . import review  # noqa: F401
 from . import anonymise  # noqa: F401
+from . import sense_check  # noqa: F401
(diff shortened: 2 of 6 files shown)


models/cfp.py
```diff
@@ -770,7 +770,7 @@ def end_date(self) -> Optional[datetime]:
 
     @property
     def potential_end_date(self) -> Optional[datetime]:
-        start = self.start_date
+        start = self.potential_start_date
```
Jonty, line 773: Vaguely concerned about side effects here given this has clearly been load-bearing, but I can't see anything obvious that was depending on the incorrect behaviour. Seems to only touch the clashfinder and attendee content.

Jonty: Actually overlaps_with is just going to be wrong if one thing has a potential time and the other doesn't. I need to think about what the correct behaviour should be there.

Jonty: Ah, context here from git digging: This was written before we had potential times, and at that point _everything_ had a potential time or it didn't.

It's still wrong but it doesn't really matter. This is fine.

Jonty (approved):

PR #1581, opened 2024-05-21
Workshop signup

### LINKED ISSUE #1443: Workshop signup system
To avoid queues of people outside workshop tents set up a lottery for tickets

Details in this pad: https://pads.orga.emfcamp.org/2024-workshop-signup?edit

supersedes #783

SamLR:
closes #1443 

I think this works, the state machines are a bit horrible and the data model isn't great. 

I'm sorry.

Instructions for seeing this:
The changes are visible on stand up of docker, running `./flask dev data` and then going to `/line-up`. You can trigger the lottery by running `./flask cfp lottery`. Going to `/cfp-review/lottery` will show the status of the tickets


templates/schedule/item.html
```diff
@@ -60,11 +61,55 @@ <h2>
...
+        <p>This {{ proposal.human_type }} does not have any tickets available via lottery.</p>
+      {% endif %}
+
+    {% elif SIGNUP_STATE == "issue-event-tickets" and proposal.requires_ticket and proposal.has_ticket_capacity() %}
+      {{ form.ticket_count.label() }}
+      {{ form.ticket_count(class_="form-control") }}
+      {{ form.get_ticket(class_="btn btn-primary") }}
+
+    {% elif SIGNUP_STATE == "issue-event-tickets" and proposal.requires_ticket and not proposal.has_ticket_capacity() %}
+      <p>This ticket is now full, if any slots become available they will be visible here.</p>
```
lukegb, line 98: Is ticket a typo here for {{proposal.human_type}}

apps/api/schedule.py
```diff
@@ -134,6 +135,36 @@ def put(self, event_id):
...
+            t.id: t
+            for t in EventTicket.query.filter_by(
+                state="entered-lottery", user_id=current_user.id
+            ).all()
+            if t.proposal.type == proposal_type
+        }
+
+        for new_rank, t_id in enumerate(new_order):
+            ticket = current_tickets[t_id]
+            ticket.rank = new_rank
```
lukegb, line 158: Tickets which don't exist in the JSON but do exist in the DB will not be reranked, so you can get everything to rank 0 by submitting this multiple times with one ticket at a time.

I don't think this helps you at all since you'll still only win once, though.

templates/cfp_review/lottery.html
```diff
@@ -0,0 +1,40 @@
...
+<h3>{{ type | capitalize }}</h3>
+<table class="table table-condensed table-striped">
+    <thead>
+        <tr>
+            <th>Title</th>
+            <th>Total tickets</th>
+            <th>Tickets via lottery</th>
+            <th>Lottery tickets issued</th>
+            <th>Tickets issued</th>
+            <th>Cancelled issued</th>
```
lukegb, line 22: Tickets cancelled?

apps/schedule/base.py
```diff
@@ -206,10 +278,72 @@ def item_current(year, proposal_id, slug=None):
...
+def cancel_event_ticket(ticket_id):
+    if current_user.is_anonymous:
+        return redirect(url_for("users.login", next=url_for("schedule.event_tickets")))
+
+    ticket = EventTicket.query.get_or_404(ticket_id)
+
+    if ticket.user != current_user:
+        abort(401)
+
+    ticket.cancel_and_update_ranks()
```
lukegb, line 340: Does this method exist?

apps/base/dev/tasks.py
```diff
@@ -35,6 +38,25 @@ def fake_data():
...
+            db.session.add(FeatureFlag(feature=flag, enabled=True))
+
+    signup_state = SiteState.query.filter_by(name="signup_state").one_or_none()
+    if not signup_state:
+        db.session.add(SiteState("signup_state", "issue-lottery-tickets"))
+    else:
+        signup_state.state = "issue-lottery-tickets"
+
+    db.session.commit()
+    db.session.flush()
```
marksteward, line 54: There are a few examples of this pattern. You shouldn't ever need to flush immediately after a commit, was something going wrong?

models/event_tickets.py
```diff
@@ -0,0 +1,190 @@
...
+        self.state = new_state
+        return self
+
+    def issue_codes(self):
+        # Now generate the ticket_codes
+        # These are in no way cryptographically secure etc but 1 in 308m should
+        # be low enough odds for guessing.
+        codes = []
+        for i in range(self.ticket_count):
+            codes.append("".join(choices(ascii_lowercase, k=6)))
```
marksteward, line 121: If this is going to be used, could you use something like `safechars = "2346789BCDFGHJKMPQRTVWXY` to make it less likely to make a pronounceable word?

apps/cfp/event_tickets_lottery.py
```diff
@@ -0,0 +1,126 @@
...
+                    ticket.won_lottery_and_cancel_others()
+                    winning_tickets.append(ticket)
+                    tickets_remaining -= ticket.ticket_count
+                else:
+                    ticket.lost_lottery()
+                db.session.commit()
+
+            proposal_capacities[proposal.id] = tickets_remaining
+
+    db.session.flush()
```
marksteward, line 97: This one's also redundant.

apps/cfp/event_tickets_lottery.py
```diff
@@ -0,0 +1,126 @@
...
+    winning_tickets = []
+
+    app.logger.info(f"Found {len(ticketed_proposals)} proposals to run a lottery for")
+    # Lock the lottery
+    signup = SiteState.query.get("signup_state")
+    if not signup:
+        raise Exception("'signup_state' not found.")
+
+    # This is the only state for running the lottery
+    signup.state = "run-lottery"
```
marksteward, line 61: I think this should check the state of the lottery before doing this, otherwise you risk re-entry.

apps/cfp/event_tickets_lottery.py
```diff
@@ -0,0 +1,126 @@
...
+    db.session.commit()
+    db.session.flush()
+    refresh_states()
+
+
+    max_rank = db.session.query(func.max(EventTicket.rank)).scalar() + 1
+    # Cache this locally so we're not hitting the db and not having to flush() etc.
```
marksteward, line 68: Ahh just saw this. `flush` doesn't mean reload from the DB, that's [`expire_all`](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.expire_all). `flush` does the opposite - it sends pending changes to the DB as SQL, so you can be sure that rows are locked by things like `update` from that point on, and can't be changed from under you. Might need to re-read this now.

marksteward: And you shouldn't really ever need to call `expire_all` fwiw

apps/cfp/event_tickets_lottery.py
```diff
@@ -0,0 +1,126 @@
...
+                continue
+
+            for ticket in tickets_for_round:
+                if ticket.ticket_count < tickets_remaining:
+                    ticket.won_lottery_and_cancel_others()
+                    winning_tickets.append(ticket)
+                    tickets_remaining -= ticket.ticket_count
+                else:
+                    ticket.lost_lottery()
+                db.session.commit()
```
marksteward, line 93: I think it'll be better to not call `.commit` in either of these places. It's preferable to have all these changes go through in a single transaction, so they either all work or none of them. As a bonus we get a single transaction in the `transaction` table that ties all the changes together.

There's still the potential for some inconsistency if someone's resubmitting their tickets all the time, and the locking we gain from that actually helps prevent raciness.

marksteward: The major effect here (other than a likely big speedup) is that we eventually end up loading all active tickets into memory. That might be quite a bit, so I want to run a sanity check on it.

apps/cfp/event_tickets_lottery.py
```diff
@@ -94,13 +96,11 @@ def run_lottery(ticketed_proposals):
...
             proposal_capacities[proposal.id] = tickets_remaining
 
-    db.session.flush()
     app.logger.info(
         f"Issued {len(winning_tickets)} winning tickets over {lottery_round} rounds"
     )
 
     signup.state = "pending-tickets"
-    db.session.commit()
     db.session.flush()
```
marksteward, line 104: Sorry, this one *should* be a commit.

SamLR: good catch :)

PR #1593, opened 2024-05-23
Remove CalendarSource and external events

marksteward:
This touches a surprising amount of stuff.

I haven't bothered removing prefixes like `proposal_favourites`, although that might be a nice follow-up commit. I couldn't see from a glance anything else that significantly simplifies in the schedule.


lukegb:
We can probably yeet https://github.com/emfcamp/Website/blob/ffa212d1e235ecc51d18c5e1fd094d29e15fc92f/templates/about/workshop-villages.md?plain=1#L58 as well

PR #1594, opened 2024-05-23
Updates to some internal phone numbers

webmariner:
Pending an important numbers section on phones.emfcamp.org, but wasn't sure if numbers for HQ or Site Manager are still the same as mentioned in other content in this repo.


lukegb:
@sammachin do you know what phone numbers are allocated for HQ/site manager?

sammachin:
I'm not sure we want the HQ and Site Manager numbers to be puiblically listed though, aside from emergencies to First Aid the first port of call to attendees should be the Info desk I believe.
HQ and Site manager numbers are on the contacts stickers that we will give to orga/volunteers

sammachin:
Also all those external numbers (01908) are outdated and should be replaced with 0333 numbers,

Overall I'm not sure we should have numbers on this page anuwauy its just another place to update each time

jellybob:
This is an update to the information page for volunteers, so I think it's fine. Probably worth emphasising in the page that people should not typically be calling HQ/site manager directly though.

sammachin:
source of truth for external numbers is the google doc in the team folder under Phones

webmariner:
Have made a few updates based on discussion above. Would just need the external number for emergency first aid I think.

marksteward (approved):

PR #1603, opened 2024-05-23
Maybe fix overlap check

marksteward:
I think this fixes the issue reported earlier today but I'm struggling to test it.


Jonty:
It might fix the exception, but I think it's still wrong because https://github.com/emfcamp/Website/pull/1603/files#diff-d84e4c839b84858f038ac30d4a90c950142ae6f1abcfb5a46017b5625cd8419dL746 means that if the other proposal starts before `self` it won't be detected as overlapping.

marksteward:
I've just matched the other overlap logic. Is that not how it should work?

marksteward:
Oh nm, I see what you mean

PR #1612, opened 2024-05-24
schedule/feeds: make json feed schema compatible

Kunsi:


apps/schedule/feeds.py
```diff
@@ -46,12 +49,146 @@ def schedule_json(year):
     return Response(json.dumps(schedule), mimetype="application/json")
 
 
+@schedule.route("/schedule/schedule-<int:year>.json")  # TODO validate url with upstream
```
Kunsi, line 52: Maybe we can just use `/schedule/<int:year>.json` and merge this pull request after this years EMF?

apps/schedule/feeds.py
```diff
@@ -32,26 +35,150 @@ def _format_event_description(event):
...
 def schedule_json(year):
     if year != event_year():
         return feed_historic(year, "json")
 
-    if not feature_enabled('SCHEDULE'):
+    if not feature_enabled("SCHEDULE"):
         abort(404)
 
-    schedule = [_convert_time_to_str(p) for p in _get_scheduled_proposals(request.args)]
+    if request.args.get('format') == 'array':
```
Kunsi, line 45: Also feels wrong to do it like this. I don't know. Feedback please! (I'm happy to expose the schema-compatible json on another endpoint, i just have no clue what would be a good name)

marksteward: I think having this exposed at /schedule-c3voc.json would be preferable - it's then clear what format it is, and it's better to return different formats based on path rather than a querystring parameter.

apps/schedule/feeds.py
```diff
@@ -32,26 +35,150 @@ def _format_event_description(event):
...
+            "day_start": day_dt.replace(hour=4, minute=0, second=0).isoformat(),
+            "index": day,
+            "rooms": {},
+        }
+        for room in rooms:
+            day_schedule["rooms"][room] = []
+            for proposal in schedule:
+                if proposal.scheduled_venue.name != room:
+                    # TODO find a better way to do that
+                    continue
```
marksteward, line 109: You can do this with `order_by(Proposal.scheduled_venue_id, Proposal.scheduled_time)` and then `itertools.groupby`.

apps/schedule/feeds.py
```diff
@@ -32,26 +35,150 @@ def _format_event_description(event):
...
+            Proposal.scheduled_venue_id.isnot(None),
+            Proposal.scheduled_duration.isnot(None),
+        )
+        .order_by(Proposal.scheduled_time)
+        .all()
+    )
+
+    duration_days = ceil((event_end() - event_end()).total_seconds / 86400)
+
+    rooms = [proposal.scheduled_venue.name for proposal in schedule]
```
marksteward, line 72: See below but this should be a set()

apps/schedule/feeds.py
```diff
@@ -32,26 +35,150 @@ def _format_event_description(event):
...
+                }
+                links.discard(None)
+                links.discard("")
+                day_schedule["rooms"][room].append(
+                    {
+                        "abstract": None,  # The proposal model does not implement abstracts
+                        "attachments": [],
+                        "date": event_tz.localize(proposal.start_date).isoformat(),
+                        "description": proposal.description,
+                        "do_not_record": False if proposal.may_record else True,
```
marksteward, line 124: We've replaced this with `video_privacy` (60bdf13)

apps/schedule/feeds.py
```diff
@@ -32,26 +35,150 @@ def _format_event_description(event):
...
+                        "room": room,
+                        "slug": "emf{}-{}-{}".format(
+                            event_year(),
+                            proposal.id,
+                            proposal.slug,
+                        ),
+                        "start": event_tz.localize(proposal.start_date).strftime("%H:%M"),
+                        "subtitle": None,
+                        "title": proposal.display_title,
+                        "track": None,  # TODO does emf have tracks?
```
marksteward, line 149: Not currently, although we may add it in the future.

marksteward (changes requested):

apps/schedule/feeds.py
Kunsi: i changed it to `/schedule/frab-<year>.json` because that's where the json export originates from.

Kunsi:
I re-set this to draft state because the machine i used to test this on is no longer available, and i don't have any other machines running docker atm

apps/schedule/schedule_json.py
```diff
@@ -0,0 +1,128 @@
...
+                            event_year(),
+                            proposal.id,
+                            proposal.slug,
+                        ),
+                        "start": event_tz.localize(proposal.start_date).strftime("%H:%M"),
+                        "subtitle": None,
+                        "title": proposal.display_title,
+                        # Contrary to the infobeamer frab module, the json module does not allow users to set colours
+                        # for tracks themselves. It instead relies on the schedule itself to provide those colours.
+                        "track": None,
```
Kunsi, line 117: We could go ahead and hard-code a mapping `proposal.type` to colour, but i'm not sure if that's a good idea

apps/schedule/schedule_json.py
wlcx, line 1: I'd suggest `schedule_frab_json.py`

apps/schedule/schedule_json.py
wlcx: I think we can leave it for now

apps/schedule/schedule_json.py
```diff
@@ -0,0 +1,128 @@
...
+                        "language": "en",
+                        "links": sorted(links),
+                        "persons": [
+                            {
+                                "name": name.strip(),
+                                "public_name": name.strip(),
+                            }
+                            for name in (proposal.published_names or proposal.user.name).split(",")
+                        ],
+                        "recording_license": "CC BY-SA 3.0",
```
wlcx, line 105: We're at 4.0 now - see #1643 

Would probably be good to set this in a single place. I expect a constant or config item would be fine!

wlcx (changes requested):
Good inclusion I think, thanks!

I wonder if we should be generating the data for this in one place though, given this is all the same stuff that's in the frab xml exporter. Removes the possiblity of accidentally having the two not agree!

apps/schedule/feeds.py
```diff
@@ -56,12 +57,51 @@ def schedule_json(year):
...
+        mimetype="application/json",
+    )
+
+
 @schedule.route("/schedule/<int:year>.frab")
 def schedule_frab(year):
+    return redirect(url_for("schedule_frab_xml", year=year), code=301)
+
+
+@schedule.route("/schedule/frab-<int:year>.xml")
```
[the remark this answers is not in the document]
Kunsi: Yeah, json and xml should have the same filename syntax.

apps/schedule/schedule_json.py
Kunsi: Would propose i'll create a separate PR for that

Kunsi:
If that's okay with you, i'd like to merge this for now, and then i'll create a follow-up PR to merge the two frab exporters into one.

apps/base/tasks_export.py
```diff
@@ -131,8 +131,13 @@ def export_db(table):
         return
 
     with app.test_client() as client:
-        for file_type in ["frab", "json", "ics"]:
-            url = f"/schedule/{year}.{file_type}"
+        for file_type, file_url in (
+            ("frab", f"{year}.frab.xml"),
+            ("frab_json", f"{year}.frab.json"),
```
marksteward, line 136: I feel like this should be `frab.json` too

Issue #1590, opened 2024-05-22
Cannot set "may record" in the admin interface on performances


Jonty:
There are performances we actually might want to record and right now we can't set it

PR #1626, opened 2024-05-24
Allow recording permission for performances

webmariner:
Fix for issue #1590

Updated the admin form and proposer form to show the checkbox for allowing recording for performances, not just talks as before. Tested both locally and ran the unit tests, all good.


webmariner:
Sorry, didn't mean to include the previous commits in my local copy of main. Will retry in a sec...

PR #1627, opened 2024-05-24
Fix for issue #1590 (Allow recording permission for performances)

webmariner:
Fix for issue https://github.com/emfcamp/Website/issues/1590

Updated the admin form and proposer form to show the checkbox for allowing recording for performances, not just talks as before. Tested both locally and ran the unit tests, all good.


lukegb:
We probably shouldn't show them the needs_laptop field though

marksteward (approved):

PR #1640, opened 2024-05-25
Initial cleanup of ages

gbaman:
First pass for only ages not currently in use for 2024. Goal being to remove other ages for 2026.

Removes
- 3+
- 4+
- 9+
- 13+
- 15+

As there are no workshops currently using these values.


marksteward:
Let's revisit this after the event, there might need to be a separate field for finalised talks.

PR #1649, opened 2024-05-25
Link to the ticket lottery about page from the schedule

nomis:
Make it easier to find the ticket lottery about page from the schedule
without having to "view your workshop tickets", which you wouldn't do
until after you've signed up to a lottery you know nothing about.
    
The "View" buttons have to be moved down otherwise putting text below
them looks odd.

Add missing space and adjust punctuation on the ticket lottery page.


marksteward:
Rebased and merged as bc970db. Thanks for this!

PR #1656, opened 2024-05-25
Add an iCal feed for volunteer shifts

edk0:


jellybob (approved):

jellybob (changes requested):
Tests are failing on mypy, I'm not entirely clear why on first glance.

edk0:
It looks like mypy just doesn't understand that mapped types are FromClauses. I pushed a change to ignore the error but I'm open to other ideas.

jellybob (approved):

PR #1658, opened 2024-05-26
Update delivery address instructions

andrewgodwin:
Remove Castle from the address and add EMF at the front to aid delivery drivers.


jellybob (approved):

edwinbalani:
Left wondering if the delivery notification I got 30mins ago has any connection to this 😅

PR #1662, opened 2024-05-26
Add /arcade and program

joenash:
This PR adds /arcade, and subpages for the content.
I copied the structure and generation of /about because having `.md` files for the entries appealed to me, but it might have been a slightly jank way of achieving this, open to suggestions. The `<details>` tag to make the nav usable might need some styling work.


marksteward:
Rebased and merged as d368da7

PR #1669, opened 2024-05-27
Add polyfill for Object.groupBy to schedule

lukegb:
Object.groupBy is from 2024, so for compatibility with older browsers let's pull in some polyfills.


lukegb:
Argh, don't merge this yet, ordering error :p

lukegb:
OK, tested in Firefox 118, and it works.

russss (approved):

PR #1671, opened 2024-05-27
Add a polyfill to make schedule viewable on old browsers without Object.groupBy

EdwardBetts:
See https://mastodon.radio/@m0rpk/112513036951002978 for bug report


russss:
You were beaten by #1669!

PR #1675, opened 2024-05-27
Replace may_record with video_privacy tri-state

marksteward:
We now have three states for Proposal.video_privacy:
 - `public` (automatically stream and publish)
 - `review` (no stream, proposer review before publish)
 - `none` (no stream, no record)

This is exposed to users in as much as they can see the state and change away from it if it's set to `review`.

Admins and heralds can set it to any state.


marksteward:
Rebased and merged as 60bdf13

PR #1687, opened 2024-05-28
There are six lightning talk slots per day

lukegb:
Per @crablab


marksteward (approved):

Jonty:
There are 8 lightning talks on Saturday, the slot is ten minutes longer

Jonty:
Unsure if it's feasible for this to look at the actual scheduled slot each day and get the duration?

PR #1689, opened 2024-05-28
Allow admins to set a user's village

### LINKED ISSUE #1684: Allow admins to add other people village admins
The correct fix is https://github.com/emfcamp/Website/issues/1526, but it would be very helpful if we could add a shitty version of this that allows admins to add people, as we are getting requests already and would like to avoid having the few of us who can login to the database doing that while sleep deprived and muddy.

edk0:
This isn't great but it might fix #1684. You have to have `admin` to see it though, which is probably wrong.

### DIFF
--- a/apps/admin/users.py
+++ b/apps/admin/users.py
@@ -3,8 +3,8 @@ import time
 from flask import render_template, redirect, flash, url_for, current_app as app, request
 from flask_login import current_user
 from flask_mailman import EmailMessage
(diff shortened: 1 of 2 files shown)


marksteward (approved):

marksteward:
It works, thanks for this!

marksteward:
Rebased and merged as 8443abb

PR #1719, opened 2024-06-01
Add content note and video privacy information to feeds

edk0:
It's been annoying me that these aren't there

### DIFF
--- a/apps/schedule/feeds.py
+++ b/apps/schedule/feeds.py
@@ -42,7 +42,7 @@ def _format_event_description(event):
         if event["map_link"]:
             venue_str = f'{venue_str} ({event["map_link"]})'
         footer_block.append(f'Venue: {venue_str}')
-    if event["type"] in ("talk", "lightningtalk", "performance") and event["video_privacy"] != "public":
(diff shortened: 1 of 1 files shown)


apps/schedule/feeds.py
```diff
@@ -35,10 +42,12 @@ def _format_event_description(event):
         if event["map_link"]:
             venue_str = f'{venue_str} ({event["map_link"]})'
         footer_block.append(f'Venue: {venue_str}')
+    if event["type"] in ("talk", "lightning talk", "performance") and event["video_privacy"] != "public":
```
lukegb, line 45: Is the type actually `lightning talk` at this point, or is it `lightningtalk`.

marksteward, line 45: Should be `lightningtalk`, yep

marksteward (changes requested):

edk0:
Lightning talks are untestable here and I only really added them because they made conceptual sense; they don't appear in the schedule afaik. Also I can't test performances because I can't make them recordable locally, did we do that via the db?

Anyway I fixed the type thing in theory, and it still looks ok for talks.

PR #1724, opened 2024-06-04
More diversity questions

SamLR:
Add disability & sexuality questions from

https://becoming-a-teacher.design-history.education.gov.uk/apply-for-teacher-training/changing-how-we-ask-about-disabilities-and-health-conditions/

and

https://www.ons.gov.uk/census/censustransformationprogramme/questiondevelopment/sexualorientationquestiondevelopmentforcensus2021


marksteward:
Is sexuality interesting or useful at all?

marksteward:
I'm not sure how useful it is for people with autism, ADHD, HIV, diabetes, or epilepsy to tell us. This feels more like a list of things that might be disabilities if untreated than things that are useful for tracking diversity, especially when we say "or health conditions".

russss:
> Is sexuality interesting or useful at all?

We want to include it for completeness, not because we suspect there are any issues there.

> I'm not sure how useful it is for people with autism, ADHD, HIV, diabetes, or epilepsy to tell us. This feels more like a list of things that might be disabilities if untreated than things that are useful for tracking diversity, especially when we say "or health conditions".

The source for that is [here](https://becoming-a-teacher.design-history.education.gov.uk/apply-for-teacher-training/changing-how-we-ask-about-disabilities-and-health-conditions/) but I do agree that it might not be the most relevant breakdown for us. For diversity stats we should be asking the broadest question which is useful, but we ideally also need more specific info on physical accessibility needs. For the time being we can punt the latter issue until later - the impetus here is that we want to put a request in the post-event email for people to fill this in.

SamLR:
I probably should have called this out in the PR itself but this form does feed data with multiple uses (which are somewhat tangential to one-another):

* Infra-planning - how should we prioritise things
* out-reach planning - which groups do we need to do more work to reach
* efficacy tracking - what has/n't worked
* sponsorship/grants work - who attends (often with specific breakdowns)
* raising the bar for others - being a good example of "these volunteers can achieve <>, you can too"

These are all broadly similar but owned by different teams (e.g. content/site/stage) and have different needs WRT forward/backwards looking.

I think I have a general preference to keep to standard forms of these questions as much as possible unless there's a good reason not as it simplifies external comparison.

For 2026 I think adding a more targetted question around specific accessibility interventions on site might be useful but are folks broadly happy with these questions as is?

apps/common/forms.py
```diff
@@ -42,6 +42,26 @@ class Meta(FlaskForm.Meta):
 AGE_VALUES = ("0-15", "16-25", "26-35", "36-45", "46-55", "56-65", "66+")
 AGE_CHOICES = tuple(OPT_OUT + [(v, v) for v in AGE_VALUES])
 
+SEXUALITY_VALUES = ("straight-or-heterosexual","gay-or-lesbian","bisexual","other")
```
Kunsi, line 45: Can we do `bi-or-pansexual` here? ("bi" as in "more than one specific gender", "pan" as in "don't care about gender")

russss (approved):
I've sorted this mostly to my satisfaction now, including a boatload of CSS changes to make it look a bit less ugly.

I'd like to put on record that I'm disappointed that this request was dumped on the web team so late, resulting in people doing work on these changes *during EMF* when they could have been enjoying the event. It's also added to my substantial post-event workload because I didn't think it was right to put out a poorly-designed, inaccessible form. (It's probably still not perfect.)

I don't think this will collect a representative amount of data on these new questions this year, and we will attach a note to the data to this effect. I also can't rule out that we will end up changing the wording of these questions for next time once we've given the subject the consideration it deserves.

PR #1730, opened 2024-06-05
Update swap-shop.md - to include specfic notice telling people not to bring anymore radioactives 

Johnr24:


Johnr24:
Making this pull request, because after EMF2024 I think we do need clarification, 
this could also be restructured so that asbestos and radioactives are in the same catch all of things which are fine when left alone and not broken apart, but bad if broken apart and absorbed by the body, 

(sorry wording is bad, am not a writer)

Johnr24:
anything containing something that when broken apart could be lethal to a human and the environment, eg, ionisation chambers,smoke-alarms & other things containing asbestos or radioactives

perhaps would be a better wording,

Johnr24:
also I am not doing this for shit-posting or memes purposes, this is intended to be a place ~~(of honour)~~ for serious discussion into how to best word this guide so people do not bring those sort of things again, 
Q mentions it well here - 
https://glauca.space/@q/112600621173302072

Words to the effect of "do not bring items that someone who does not understand the risks may walk away with", could also work.

SamLR:
hi @Johnr24, thanks for the PR & sorry for the slow response. 

Think this looks fine

Johnr24:
> hi @Johnr24, thanks for the PR & sorry for the slow response.
> 
> Think this looks fine

Thanks! 
I've just remerged with the most recent commits, this should be ready to go once approved!

SamLR (approved):

Issue #1259, opened 2022-09-01
Email speakers when videos are published


russss:
It would be good to email speakers a link to their talk video when it's published. Some thoughts:

* We should probably abstract this somewhere, so we can set talk videos through an internal API or through a command-line tool. (We have some existing video-matching code [here](https://github.com/emfcamp/Website/blob/main/apps/base/tasks_videos.py) but it's a bit brittle).
* We should only email the speaker the first time this happens (videos may not be uploaded to all video services simultaneously).
* We should probably also email speakers when the "video lost" flag is set (#1258)

jayaddison:
Some ad-hoc requirements gathering from IRC:

- [ ] On the EMF talk/proposal view ([example](https://www.emfcamp.org/schedule/2022/139-a-brief-history-of-time-zones-and-daylight-saving-time)), add a list of hyperlinks to video URLs below the embedded video object.
- [ ] When any non-empty URL is first configured for a `Proposal`, send a one-time email to the speaker with a link to the EMF talk/proposal view URL.

fkusei:
I'm currently working on a "generic webhook" feature for our publishing scripts which would maybe help here: https://github.com/voc/voctopublish/pull/109 (feedback welcome)

(Requires changing the API on the website though)

wlcx:
I'm not 100% on emailing the speaker in the case of missing recording - there may be more nuance to it than a generic catch all "something went wrong" message, and in any case I think I'd rather email them with a more personal message and a name attached rather than a "cold" website email.

Of course, that's just me and it's probably better they get something rather than nothing - so maybe the best course is to proceed but also send them a more human email? Just thinking out loud.

crablab:
Agreed. I don't think we should automatically email people with lost recordings. 

We should probably send an alert to the CfP channel instead, so someone from the Content team can reach out directly and discuss with the speaker.

That assumes this is a relatively rare occurrence?

russss:
Historically it has not really been a rare occurrence but I'm hopeful we won't have any lost recordings this year, so I'm happy to skip that for the moment at least.

russss:
Some goalpost-moving, sorry:

* It looks like we will need a way of sending the emails after the URLs are set. I wonder if it would be best to have this as a command-line task, but we'd need to store the email status somewhere. We could maybe add a "video-published" state to the state machine?
* We may not make YouTube videos public at the same time that they're uploaded. So I think it's best to make this trigger on the `c3voc_url` only for the moment.

[...]

PR #1740, opened 2024-06-17
[proposal view] Add video recording host hyperlinks.

jayaddison:
As part of #1259 we're going to email speakers when their edited video recordings from EMF are first published, and in the emails we're going to link to the proposal view page on the emfcamp.org website.

To make it easier for speakers and viewers to locate the origin hosted copies of the recordings, we should include direct links to each platform (currently CCC, YouTube) on the proposal view.

Note: untested because I don't currently have a functional local dev env (any help testing this is appreciated!)

### DIFF
--- a/templates/schedule/historic/item.html
+++ b/templates/schedule/historic/item.html
@@ -64,7 +64,7 @@
     <strong>Video Hosts:</strong>
     <ul>
     {% if event.video.ccc %}
-      <li>View this video <a href="{{ event.video.ccc }}">on the Chaos Computer Club's website</a>.</li>
+      <li>View this video <a href="{{ event.video.ccc }}">on media.ccc.de</a>.</li>
     {% endif %}
     {% if event.video.youtube %}
       <li>View this video <a href="{{ event.video.youtube }}">on YouTube</a>.</li>
--- a/templates/schedule/historic/item.html
+++ b/templates/schedule/historic/item.html
@@ -60,8 +60,16 @@
           <button class="youtube-wait-for-click-button">Continue</button>
         </div>
       </div>
-      <p>View this video <a href="{{ event.video.youtube }}">on YouTube</a>.</p>
     {% endif %}
+    <strong>Video Hosts:</strong>
+    <ul>
+    {% if event.video.ccc %}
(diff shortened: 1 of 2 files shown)


templates/schedule/item.html
```diff
@@ -135,8 +135,16 @@ <h4>Video</h4>
     <iframe src="{{proposal.youtube_url|replace('youtube.com/watch?v=', 'youtube.com/embed/')}}"
         width="100%" height="500px" frameborder="0" 
         allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
-    <p>View this video <a href="{{ proposal.youtube_url }}">on YouTube</a>.</p>
   {% endif %}
+  <p>Direct links to the host(s) of this video recording can be found below:</p>
```
jayaddison, line 139: This sentence has a verbose/technical tone to it - it doesn't read naturally.

Originally I thought about simply using a section heading of `External Links` -- but I didn't want to imply that the embedded video on the page is hosted on the EMF website (because it isn't).

jayaddison, line 139:
```suggestion
  <p>Video Hosts</p>
```

jayaddison:
Based on local testing, this appears to be working fine.  The layout / design doesn't seem ideal to me, so perhaps someone could suggest improvements, but it does function.

**Screenshot (with empty video object)**
image

templates/schedule/historic/item.html
```diff
@@ -60,8 +60,16 @@ <h3>Video</h3>
           <button class="youtube-wait-for-click-button">Continue</button>
         </div>
       </div>
-      <p>View this video <a href="{{ event.video.youtube }}">on YouTube</a>.</p>
     {% endif %}
+    <strong>Video Hosts:</strong>
+    <ul>
+    {% if event.video.ccc %}
+      <li>View this video <a href="{{ event.video.ccc }}">on the Chaos Computer Club's website</a>.</li>
```
Jonty, line 67: This should probably be "on media.ccc.de"

russss (approved):

PR #1751, opened 2024-06-27
Add `pump` dependency, missing after #1735

lukegb:
Oops.


jayaddison (approved):
Looks good to me :+1: 

(this prompted me to do a quick manual inspection to check that all of the dependencies imported by the `gulpfile.js` are listed in `package.json`; with this in place, indeed they all are)

PR #1768, opened 2024-08-27
cfp: Add Video download url

wlcx:
This is used to store a canonical url for the video recording of a CFP item, used to automate release of the recording by means of the schedule frab/json.
This is useful in particular when releasing videos via media.ccc.de which can import our frab schedule and automatically download the videos from the listed URLs, but it could also be used to automate youtube uploads, etc.
We also add the ability to set this via the proposal api endpoint.

I'm not entirely convinced this should this be implemented in this way though:
- Arguably this information is only of use to known automated tools consuming the schedule - perhaps it should require a url flag or something to include it? 
- There are perhaps some minor security concerns around making publicly available the server being used to host the finished recordings
- These URLs would likely be ephemeral, going away at some unspecified time post-event once video team are done releasing. There's not much use them being included in the schedule export and preserved for all time, for instance.


apps/api/schedule.py
```diff
@@ -17,10 +17,16 @@ def _require_video_api_key(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         auth_header = request.headers.get("authorization", None)
-        if not auth_header or not auth_header.startswith("Bearer "):
+        if not auth_header:
+            abort(401)
+
+        if auth_header.startswith("bearer "):
+            bearer_token = auth_header.removeprefix("bearer ")
```
marksteward, line 24: The contents of the Authorization header is case-sensitive, so "bearer" should be wrong. Can you add a comment to explain why this is needed?

jayaddison:
Same question about the contents of the authorization header, but apart from that: looks good to me :+1:

jayaddison:
> There are perhaps some minor security concerns around making publicly available the server being used to host the finished recordings

I don't know whether this would be necessary -- it depends on the workflows and server URL schemes involved -- but we could perhaps assert that the stored values don't contain any authentication details (`http://user:creds@example.invalid`) and that they don't contain query-strings (sometimes used to embed API tokens - `http://example.invalid/api/endoint?key=adsf`) - to prevent some easily-avoidable accidental mistakes.

wlcx:
> > Arguably this information is only of use to known automated tools consuming the schedule - perhaps it should require a url flag or something to include it?
> 
> Is the schedule format adjusted here only presented through the `/schedule/<year>.frab` endpoint?

Nope, json too

apps/api/schedule.py
wlcx: I added it because I wasted 5 minutes scratching my head why my curl request was giving a 401 ;-)

I went looking for consensus online, always a mistake.. honestly, it seems like a bit of a mess: https://sgeb.io/posts/fix-go-oauth2-case-sensitive-bearer-auth-headers/

Given we're not implementing oauth here I figured it was a convenience and didn't worry about it too much.

wlcx:
Closing - will generate this separately since it's a bit single-purpose and doesn't really need to live in the website.

apps/api/schedule.py
marksteward: Ahh, fair, `# make it easier with curl` would be fine then

PR #1770, opened 2024-08-31
api.schedule: add C3VOCPublishingWebhook

Kunsi:
This will allow c3voc to send us video information automatically upon release of videos, as documented here: https://github.com/voc/voctopublish/blob/main/voctopublish/api_client/webhook_client.py#L28-L56

### DIFF
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -148,9 +148,7 @@ class UpdateLotteryPreferences(Resource):
 
         current_tickets = {
             t.id: t
-            for t in EventTicket.query.filter_by(
-                state="entered-lottery", user_id=current_user.id
-            ).all()
+            for t in EventTicket.query.filter_by(state="entered-lottery", user_id=current_user.id).all()
             if t.proposal.type == proposal_type
         }
 
@@ -190,14 +188,17 @@ class ProposalC3VOCPublishingWebhook(Resource):
                 abort(403, message="The request referenced a non-master video edit, and has been denied.")
 
             if conference != f"emf{event_year()}":
-                abort(422, message="The request did not reference the current event year, and has not been processed.")
+                abort(
+                    422,
+                    message="The request did not reference the current event year, and has not been processed.",
+                )
 
             proposal = Proposal.query.get_or_404(proposal_id)
 
             if payload["voctoweb"]["enabled"]:
                 if payload["voctoweb"]["frontend_url"]:
                     c3voc_url = payload["voctoweb"]["frontend_url"]
-                    if not c3voc_url.startswith('https://media.ccc.de/'):
+                    if not c3voc_url.startswith("https://media.ccc.de/"):
                         abort(406, message="voctoweb frontend_url must start with https://media.ccc.de/")
                     app.logger.info(f"C3VOC webhook set c3voc_url for {proposal.id=} to {c3voc_url}")
                     proposal.c3voc_url = c3voc_url
@@ -207,7 +208,9 @@ class ProposalC3VOCPublishingWebhook(Resource):
                     # as well. We do not explicitely set 'video_recording_lost'
                     # here because the video might only need fixing audio or
                     # such.
-                    app.logger.warning(f"C3VOC webhook cleared c3voc_url for {proposal.id=}, was {proposal.c3voc_url}")
+                    app.logger.warning(
+                        f"C3VOC webhook cleared c3voc_url for {proposal.id=}, was {proposal.c3voc_url}"
+                    )
                     proposal.c3voc_url = ""
 
             if payload["youtube"]["enabled"]:
@@ -215,7 +218,7 @@ class ProposalC3VOCPublishingWebhook(Resource):
                     # Please do not overwrite existing youtube urls
                     if not proposal.youtube_url:
                         youtube_url = payload["youtube"]["urls"][0]
-                        if  not youtube_url.startswith('https://www.youtube.com/watch'):
+                        if not youtube_url.startswith("https://www.youtube.com/watch"):
                             abort(406, message="youtube url must start with https://www.youtube.com/watch")
                         # c3voc will send us a list, even though we only have one
                         # video.
@@ -228,10 +231,14 @@ class ProposalC3VOCPublishingWebhook(Resource):
                         app.logger.warning(
                             "C3VOC webhook sent youtube urls update without referencing the previously stored value. Ignoring."
                         )
-                        app.logger.debug(f"{proposal.id=} {payload['youtube']['urls']=} {proposal.youtube_url=}")
+                        app.logger.debug(
+                            f"{proposal.id=} {payload['youtube']['urls']=} {proposal.youtube_url=}"
+                        )
                 else:
                     # see comment at c3voc_url above
-                    app.logger.warning(f"C3VOC webhook cleared youtube_url for {proposal.id=}, was {proposal.youtube_url}")
+                    app.logger.warning(
+                        f"C3VOC webhook cleared youtube_url for {proposal.id=}, was {proposal.youtube_url}"
+                    )
                     proposal.youtube_url = ""
 
             db.session.add(proposal)
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -216,10 +216,10 @@ class ProposalC3VOCPublishingWebhook(Resource):
             if payload["youtube"]["enabled"]:
                 if payload["youtube"]["urls"]:
                     # Please do not overwrite existing youtube urls
+                    youtube_url = payload["youtube"]["urls"][0]
+                    if not youtube_url.startswith("https://www.youtube.com/watch"):
+                        abort(406, message="youtube url must start with https://www.youtube.com/watch")
                     if not proposal.youtube_url:
-                        youtube_url = payload["youtube"]["urls"][0]
-                        if not youtube_url.startswith("https://www.youtube.com/watch"):
-                            abort(406, message="youtube url must start with https://www.youtube.com/watch")
                         # c3voc will send us a list, even though we only have one
                         # video.
                         app.logger.info(f"C3VOC webhook set youtube_url for {proposal.id=} to {youtube_url}")
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -196,7 +196,7 @@ class C3VOCPublishingWebhook(Resource):
 
         proposal = Proposal.query.get_or_404(proposal_id)
 
-        if payload["voctoweb"]["enabled"]:
+        if payload["voctoweb"]["enabled"] and payload["voctoweb"]["frontend_url"]:
             proposal.c3voc_url = payload["voctoweb"]["frontend_url"]
             proposal.video_recording_lost = False
         else:
@@ -206,7 +206,7 @@ class C3VOCPublishingWebhook(Resource):
             # such.
             proposal.c3voc_url = ""
 
-        if payload["youtube"]["enabled"]:
+        if payload["youtube"]["enabled"] and payload["youtube"]["urls"]:
             # c3voc will send us a list, even though we only have one
             # video.
             proposal.youtube_url = payload["youtube"]["urls"][0]
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -182,62 +182,62 @@ class ProposalC3VOCPublishingWebhook(Resource):
         try:
             conference = payload["fahrplan"]["conference"]
             proposal_id = payload["fahrplan"]["id"]
-        except KeyError:
-            abort(422)
 
-        if not payload["is_master"]:
-            # c3voc *should* only send us information about the master
-            # encoding getting published. Aborting early ensures we don't
-            # accidentially delete video information from the database.
-            abort(403)
+            if not payload["is_master"]:
+                # c3voc *should* only send us information about the master
+                # encoding getting published. Aborting early ensures we don't
+                # accidentially delete video information from the database.
+                abort(403, message="The request referenced a non-master video edit, and has been denied.")
 
-        if conference != f"emf{event_year()}":
-            abort(422)
+            if conference != f"emf{event_year()}":
+                abort(422, message="The request did not reference the current event year, and has not been processed.")
 
-        proposal = Proposal.query.get_or_404(proposal_id)
+            proposal = Proposal.query.get_or_404(proposal_id)
 
-        if payload["voctoweb"]["enabled"]:
-            if payload["voctoweb"]["frontend_url"]:
-                c3voc_url = payload["voctoweb"]["frontend_url"]
-                if not c3voc_url.startswith('https://media.ccc.de/'):
-                    abort(406, message="voctoweb frontend_url must start with https://media.ccc.de/")
-                app.logger.info(f"C3VOC webhook set c3voc_url for {proposal.id=} to {c3voc_url}")
-                proposal.c3voc_url = c3voc_url
-                proposal.video_recording_lost = False
-            else:
-                # This allows c3voc to notify us if videos got depublished
-                # as well. We do not explicitely set 'video_recording_lost'
-                # here because the video might only need fixing audio or
-                # such.
-                app.logger.warning(f"C3VOC webhook cleared c3voc_url for {proposal.id=}, was {proposal.c3voc_url}")
-                proposal.c3voc_url = ""
-
-        if payload["youtube"]["enabled"]:
-            if payload["youtube"]["urls"]:
-                # Please do not overwrite existing youtube urls
-                if not proposal.youtube_url:
-                    youtube_url = payload["youtube"]["urls"][0]
-                    if  not youtube_url.startswith('https://www.youtube.com/watch'):
-                        abort(406, message="youtube url must start with https://www.youtube.com/watch")
-                    # c3voc will send us a list, even though we only have one
-                    # video.
-                    app.logger.info(f"C3VOC webhook set youtube_url for {proposal.id=} to {youtube_url}")
-                    proposal.youtube_url = youtube_url
+            if payload["voctoweb"]["enabled"]:
+                if payload["voctoweb"]["frontend_url"]:
+                    c3voc_url = payload["voctoweb"]["frontend_url"]
+                    if not c3voc_url.startswith('https://media.ccc.de/'):
+                        abort(406, message="voctoweb frontend_url must start with https://media.ccc.de/")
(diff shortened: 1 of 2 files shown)


Kunsi:
Tested:

```
# no json content-type
/home/kunsi➤ curl -X POST -H"Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook
{
    "message": "The server does not support the media type transmitted in the request."
}

# no json data
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.
138.22:2342/api/c3voc/publishing-webhook
{
    "message": "Failed to decode JSON object: Expecting value: line 1 column 1 (char 0)"
}

# empty json
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook --data '{}'
{
    "message": "The request was well-formed but was unable to be followed due to semantic errors."
}

# no master ticket
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook --data '{"is_master": false, "fahrplan": {"conference": "emf2024", "id": 12}}'
{
    "message": "The resource identified by the request is only capable of generating response entities which have content characteristics not acceptable according to the accept headers sent in the request."
}

# sets youtube url only
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook --data '{"is_master": true, "fahrplan": {"conference": "emf2024", "id": 14}, "voctoweb": {"enabled": false}, "youtube": {"enabled": true, "urls": ["https://www.youtube.com/watch?v=hsoiHKaqG2s"]}}'

# sets both youtube url and media.ccc.de url
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook --data '{"is_master": true, "fahrplan": {"conference": "emf2024", "id": 14}, "voctoweb": {"enabled": true, "frontend_url": "https://media.ccc.de/v/eh21-61-state-of-bahn-reisendeninformation"}, "youtube": {"enabled": true, "urls": ["https://www.youtube.com/watch?v=hsoiHKaqG2s"]}}'

# unsets youtube url
/home/kunsi➤ curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer video-api-token" http://172.19.138.22:2342/api/c3voc/publishing-webhook --data '{"is_master": true, "fahrplan": {"conference": "emf2024", "id": 14}, "voctoweb": {"enabled": true, "frontend_url": "https://media.ccc.de/v/eh21-61-state-of-bahn-reisendeninformation"}, "youtube": {"enabled": false}}'
```

apps/api/schedule.py
```diff
@@ -186,3 +237,4 @@ def get(self):
 api.add_resource(FavouriteExternal, "/external/<int:event_id>/favourite")
 api.add_resource(ScheduleMessage, "/schedule_messages")
 api.add_resource(UpdateLotteryPreferences, "/schedule/tickets/<proposal_type>/preferences")
+api.add_resource(C3VOCPublishingWebhook, "/c3voc/publishing-webhook")
```
jayaddison, line 240: To align with some existing webhook URLs (for example [this Stripe webhook endpoint](
https://github.com/emfcamp/Website/blob/888fb8b7d63c7560376bc2d5670b0e46f61fdce0/apps/payments/stripe.py#L171-L172)), and to indicate the key affected resource in the path, I'd suggest:

```suggestion
api.add_resource(ProposalC3VOCWebhookResource, "/proposal/c3voc-webhook")
```

apps/api/schedule.py
```diff
@@ -169,6 +170,56 @@ def post(self, proposal_type):
...
+            conference = payload["fahrplan"]["conference"]
+            proposal_id = payload["fahrplan"]["id"]
+        except KeyError:
+            abort(422)
+
+        if not payload["is_master"]:
+            # c3voc *should* only send us information about the master
+            # encoding getting published. Aborting early ensures we don't
+            # accidentially delete video information from the database.
+            abort(406)
```
jayaddison, line 192: Although HTTP 403 is _usually_ authorization-related, in this case I think it might be the appropriate status code here.  We're denying the edit to the resource because we don't think it's allowable given the contents of the request.

```suggestion
            abort(403)
```

Out of interest: what would non-master videos be?  (would those be in-progress drafts / edits, for example?)

apps/api/schedule.py
```diff
@@ -169,6 +170,56 @@ def post(self, proposal_type):
...
+            # such.
+            proposal.c3voc_url = ""
+
+        if payload["youtube"]["enabled"] and payload["youtube"]["urls"]:
+            # c3voc will send us a list, even though we only have one
+            # video.
+            proposal.youtube_url = payload["youtube"]["urls"][0]
+            proposal.video_recording_lost = False
+        else:
+            proposal.youtube_url = ""
```
jayaddison, line 215:
```suggestion
        if payload["youtube"]["enabled"] and payload["youtube"]["urls"]:
            if proposal.youtube_url is None:
                # c3voc will send us a list, even though we only have one video.
                proposal.youtube_url = payload["youtube"]["urls"][0]
            else:
                # log a warning?
                pass
```

Some reasoning for this suggestion:

  * If `youtube_url` is already configured for some reason, I don't think that we should overwrite or remove the existing non-empty value.
  * Although it _sort of_ makes sense to unset `video_recording_lost` here, I think it may be better to leave it unmodified (I don't know how/why it could happen that V3VOC sends a video and yet EMF thought it was lost -- but let's allow that to be discoverable if it does?).

Kunsi: @wlcx do you have opinions about unsetting `video_recording_lost` here? My assumption is "if we publish a video, the recording obviously cannot be lost", but that might be wrong?

apps/api/schedule.py
```diff
@@ -169,6 +170,56 @@ def post(self, proposal_type):
...
+
+        if payload["voctoweb"]["enabled"] and payload["voctoweb"]["frontend_url"]:
+            proposal.c3voc_url = payload["voctoweb"]["frontend_url"]
+            proposal.video_recording_lost = False
+        else:
+            # This allows c3voc to notify us if videos got depublished
+            # as well. We do not explicitely set 'video_recording_lost'
+            # here because the video might only need fixing audio or
+            # such.
+            proposal.c3voc_url = ""
```
jayaddison, line 207: I'm wondering about the use-cases for this: is this also necessary to handle situations where the content should be removed?  (some Twitter/X API events for example include events to indicate that posts should be removed downstream)

apps/api/schedule.py
Kunsi: A non-master would be something like a scaled-down version of the same video - these *should* have the same urls, but i would not vouch for that.

Regarding the http status code, i'm not sure either. My thought was "it's not acceptable here to do updates for non-master encodings", but i can also fully understand the thought of "it's not allowed to do that". Will give it a thought.

apps/api/schedule.py
```diff
@@ -169,6 +170,59 @@ def post(self, proposal_type):
...
+
+        if payload["youtube"]["enabled"] and payload["youtube"]["urls"]:
+            # only update the youtube url if we don't currently have one
+            # stored.
+            if not not proposal.youtube_url:
```
jayaddison, line 212: Careful here :)

```suggestion
            if not proposal.youtube_url:
```

apps/api/schedule.py
Kunsi: Sometimes we (c3voc) we get requests to take a video down for various reasons (accidentially published, copyright/slide fixes or such things). In this case the reference to the no-longer-existing videos should be removed from emf website as well, because we don't want to link to "this video has been removed" pages.

apps/api/schedule.py
Kunsi: you did not see nothing :D

apps/api/schedule.py
jayaddison: Ok, thanks - that makes sense; and I suppose that, to be cautious, we would also want to remove the `youtube_url` when that happens - a slight conflict with my previous suggestion not to edit existing values for that attribute.

Kunsi: i have reverted the "don't update existing youtube urls" commit. I'd wait for input of @wlcx on this as well.

apps/api/schedule.py
wlcx: > If youtube_url is already configured for some reason, I don't think that we should overwrite or remove the existing non-empty value.
Agree, although you could check for equality and squash the log message if true which should help most of the time.

> do you have opinions about unsetting video_recording_lost here?
I think it's probably fine to unset it, the whole point of it was to provide a reason as to why we didn't have video URLs published - it would be very confusing to have a message saying we lost the recording and then have the recording appear below.

jayaddison: > I think it's probably fine to unset it, the whole point of it was to provide a reason as to why we didn't have video URLs published - it would be very confusing to have a message saying we lost the recording and then have the recording appear below.

Ok, yep - now that I notice the [corresponding template block](https://github.com/emfcamp/Website/blob/888fb8b7d63c7560376bc2d5670b0e46f61fdce0/templates/schedule/item.html#L16-L20).. agreed, that would be confusing.  And it's probably not worthwile saying something like "we thought we'd lost it, but it's here, tada" - so unsetting the field does make sense.

jayaddison:
Commencing some local testing of this here (the code looks good to me)

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+
+        payload = request.get_json()
+
+        try:
+            conference = payload["fahrplan"]["conference"]
+            proposal_id = payload["fahrplan"]["id"]
+        except KeyError:
+            abort(422)
+
+        if not payload["is_master"]:
```
jayaddison, line 188:
```suggestion
        if not payload.get("is_master"):
```

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+            # c3voc *should* only send us information about the master
+            # encoding getting published. Aborting early ensures we don't
+            # accidentially delete video information from the database.
+            abort(403)
```
jayaddison, line 192:
```suggestion
            abort(403, message="The request referenced a non-master video edit, and has been denied.")
```

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+
+        if conference != f"emf{event_year()}":
+            abort(422)
```
jayaddison, line 195:
```suggestion
            abort(422, message="The request did not reference the current event year, and has not been processed.")
```

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+
+        proposal = Proposal.query.get_or_404(proposal_id)
+
+        if payload["voctoweb"]["enabled"]:
+            if payload["voctoweb"]["frontend_url"]:
+                proposal.c3voc_url = payload["voctoweb"]["frontend_url"]
```
jayaddison, line 201:
```suggestion
        if payload.get("voctoweb", {}).get("enabled") is True:
            if c3voc_url := payload["voctoweb"].get("frontend_url"):
                proposal.c3voc_url = c3voc_url
```

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+                proposal.video_recording_lost = False
+            else:
+                # This allows c3voc to notify us if videos got depublished
+                # as well. We do not explicitely set 'video_recording_lost'
+                # here because the video might only need fixing audio or
+                # such.
+                proposal.c3voc_url = ""
```
jayaddison, line 208:
```suggestion
                proposal.c3voc_url = None
```

jayaddison (commented):
Mostly nitpicks so far; here's what I've tested, more-or-less:

```bash
$ curl -XPOST http://127.0.0.1:2342/api/proposal/c3voc-publishing-webhook
{
    "message": "The server could not verify that you are authorized to access the URL requested. You either supplied the wrong credentials (e.g. a bad password), or your browser doesn't understand how to supply the credentials required."
}

$ curl -H 'Authorization: Bearer video-api-token' -XPOST http://127.0.0.1:2342/api/proposal/c3voc-publishing-webhook
{
    "message": "The server does not support the media type transmitted in the request."
}

$ curl -H 'Content-Type: application/json' -H 'Authorization: Bearer video-api-token' -XPOST http://127.0.0.1:2342/api/proposal/c3voc-publishing-webhook --data '{}'
{   
    "message": "The request was well-formed but was unable to be followed due to semantic errors."
}

$ curl -H 'Content-Type: application/json' -H 'Authorization: Bearer video-api-token' -XPOST http://127.0.0.1:2342/api/proposal/c3voc-publishing-webhook --data '{"fahrplan": {"conference": "emf2026", "id": 5}}'
# returned a slightly confusing authorization error message

$ curl -H 'Content-Type: application/json' -H 'Authorization: Bearer video-api-token' -XPOST http://127.0.0.1:2342/api/proposal/c3voc-publishing-webhook --data '{"is_master": true, "fahrplan": {"conference": "emf2026", "id": 5}}'
# returned a confusing error message about semantics
```

jayaddison:
Since we display the URLs on the relevant event pages, and since both `media.ccc.de` and `youtube.com` redirect to HTTPS by default, let's add validation of the URLs before storing them in the database.

```
                if not c3voc_url.startswith('https://media.ccc.de/'):
                    abort(403, message='Please ensure that the URL points to the media.ccc.de HTTPS service.')
```

or something along those lines

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+                    proposal.youtube_url = payload["youtube"]["urls"][0]
+                    proposal.video_recording_lost = False
+                elif proposal.youtube_url not in payload["youtube"]["urls"]:
+                    # c3voc sent us some urls, but none of them are matching
+                    # the url we have in our database.
+                    app.logger.warning(
+                        f"C3VOC webhook sent youtube urls {payload['youtube']['urls']!r}, "
+                        f"but we already have {proposal.youtube_url}. NOT "
+                        "overwriting!"
+                    )
```
jayaddison, line 225:
```suggestion
                    app.logger.warning("C3VOC webhook contained a YouTube URL update without referencing our previously-stored value.  Ignoring.")
```

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+            else:
+                # see comment at c3voc_url above
```
jayaddison, line 227:
```suggestion
                    app.logger.warning(f"C3VOC webhook cleared the YouTube URL for event_id={proposal.id}")

```

jayaddison: (edited to fix indentation...)

apps/api/schedule.py
```diff
@@ -169,6 +170,69 @@ def post(self, proposal_type):
...
+                proposal.youtube_url = ""
+
+        db.session.add(proposal)
+        db.session.commit()
+
+        return "OK", 204
```
jayaddison, line 238: Nitpick: we should be able to omit this (the default return status code is `200` / success; it looks like the `voctopublish` client code should handle that without a problem).

apps/api/schedule.py
Kunsi: Took the liberty of adding some more logging, even for the success cases.

apps/api/schedule.py
Kunsi: Yeah, but since i've built some other APIs in the past, i feel only returning "OK" warrants a `204 no content` response. If we return 200 here, we should return (parts of) the proposal object, which feels unnecessary here (noone ever looks at it).

apps/api/schedule.py
Kunsi: `ProposalResource` sets it to `""` on top of the file :/

jayaddison: Ah, drats.  Not a nullable field.  Ok, `""` it is!

apps/api/schedule.py
Kunsi: I don't like splattering code with lots of `.get()`s. If we fear c3voc might send us a webhook with a bunch of data missing, i'd rather wrap everything in `try..except: abort(400)` ...

apps/api/schedule.py
jayaddison: Mm.. my mistake; in fact it _is_ nullable: https://github.com/emfcamp/Website/blob/888fb8b7d63c7560376bc2d5670b0e46f61fdce0/migrations/versions/9483c0756933_add_proposal_urls.py#L22

I wonder if there's any importance in using `""` instead of `None` there (cc @lukegb?).

apps/api/schedule.py
Kunsi: I highly despise the `:=`-Syntax recent python versions introduced, because (in my opinion) they make the code less readable. I have implemented parts of that suggestion anyway.

apps/api/schedule.py
jayaddison: Seems reasonable to me.  Thank you for the explanation :+1:

apps/api/schedule.py
jayaddison: That seems to be the cause of a bug; I've branched that off into #1771 (not a blocker for this).

jayaddison (approved):
Looks good to me :+1: 

My two remaining nitpicks would be:

  * We don't have much `info`-level logging in the API modules of the codebase so far, so I don't know whether we should include those.
  * The `try...catch` for `KeyError` seems overbroad to me; but I'm unaware of any particular problem that would cause.

jayaddison:
(ah: and one more thing: maybe some of the line-lengths could be reduced; but again, a nitpicky detail)

jayaddison:
Another one-more-thing: we should probably add some test coverage for this.  I'm prety convinced from manually testing it that it works well, but for the longer-term, some automated coverage would be nice.  Borrowing some of the [existing test code for the `PATCH` endpoint](https://github.com/emfcamp/Website/blob/888fb8b7d63c7560376bc2d5670b0e46f61fdce0/tests/test_api_proposals.py#L37-L60) could provide a base for that.

Kunsi:
I think my tests cover all the cases we'd want to test. Let me know if i missed any.

russss (approved):

PR #1774, opened 2024-09-01
ProposalC3VOCPublishingWebhook: add option to set thumbnail as well (and test naming)

Kunsi:
... and better test naming

### DIFF
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -213,6 +213,20 @@ class ProposalC3VOCPublishingWebhook(Resource):
                     )
                     proposal.c3voc_url = None
 
+                if payload["voctoweb"]["thumb_path"]:
+                    path = payload["voctoweb"]["thumb_path"]
+                    if path.startswith("/static.media.ccc.de"):
+                        path = "https://static.media.ccc.de/media" + path[len("/static.media.ccc.de"):]
+                    if not path.startswith("https://"):
+                        abort(406, message="voctoweb thumb_path must start with https:// or /static.media.ccc.de")
+                    app.logger.info(f"C3VOC webhook set thumbnail_url for {proposal.id=} to {path}")
+                    proposal.thumbnail_url = path
+                else:
+                    app.logger.warning(
+                        f"C3VOC webhook cleared thumbnail_url for {proposal.id=}, was {proposal.thumbnail_url}"
+                    )
+                    proposal.thumbnail_url = None
+
             if payload["youtube"]["enabled"]:
                 if payload["youtube"]["urls"]:
                     # Please do not overwrite existing youtube urls
--- a/tests/test_api_c3voc.py
+++ b/tests/test_api_c3voc.py
@@ -110,7 +110,7 @@ def test_denies_request_wrong_year(client, app, proposal):
     assert rv.status_code == 422
 
 
-def test_request_none_update_none(client, app, db, proposal):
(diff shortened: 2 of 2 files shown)


russss (approved):

russss (approved):

tests/test_api_c3voc.py
```diff
@@ -241,6 +244,81 @@ def test_request_voctoweb_clears_voctoweb(client, app, db, proposal):
...
+
+
+def test_update_thumbnail_with_path(client, app, db, proposal):
+    app.config.update(
+        {
+            "VIDEO_API_KEY": "api-key",
+        }
+    )
+
+    clean_proposal(db, proposal)
```
jayaddison, line 267: Just a note that I'm attempting two cleanup refactorings locally, although running into some challenges:

  * I think we could move most of the `clean_proposal` -- which seems a mixture of unit test setup and teadown logic -- into the fixture itself, by using [`yield`](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended) instead of `return`.
  * We could _probably_ move the `VIDEO_API_KEY` into the `config/test.cfg` file -- although I guess even so we need to override it for one or two of the tests.

apps/api/schedule.py
```diff
@@ -216,6 +216,20 @@ def post(self):
                     )
                     proposal.c3voc_url = None
 
+                if payload["voctoweb"]["thumb_path"]:
+                    path = payload["voctoweb"]["thumb_path"]
+                    if path.startswith("/static.media.ccc.de"):
+                        path = "https://static.media.ccc.de/media" + path[len("/static.media.ccc.de"):]
+                    if not path.startswith("https://"):
```
jayaddison, line 223: Are other CDNs likely to be used to host these thumbnails? (just wondering about the broader URI acceptance logic in this case)

Kunsi: This URL has changed in the past (some time ago it was `cdn.c3voc.de` i think). I don't think this will change in the future though, so we could have implemented stricter checking here, you're right.

tests/test_api_c3voc.py
Kunsi: regarding `clean_proposal()`: yes. personally i'm a fan of starting every test on the same clean slate, so we should try to implement that here as well.

I have no opinions about the VIDEO_API_KEY.

jayaddison: There's a relatively nice pattern in the Sphinx unit tests where tests can have a decorator applied to them that configures some settings temporarily for the duration of a test using a dictionary object.  It keeps the declarative / test setup near the test case, but is a bit more concise (writing to `app.config` within test cases seems strange).

Where I'm stuck with the fixture teardown refactoring is that some errors are occurring when I run the tests in random order.  So something about my adjusted teardown logic must be incorrect..

I don't think any of this is essential, but I'd like to try to provide a follow-up pull request if I get it working.

jayaddison: Possible test adjustments opened at #1776 - the fixture teardown problem was because I hadn't noticed the `module`-level scoping of the fixture.

apps/api/schedule.py
jayaddison: Thanks, OK!  For consistency and safety, I think it makes sense to check the complete domain in that case too.

PR #1802, opened 2025-05-31
Export village information

### LINKED ISSUE #1548: Villages export for next cycle
At the end of each event, it would be helpful to have a way to export at least the following information about villages:

- Name
- Key person of contact
- Contact details
- Number of Attendees

This will help us have a contact list to send vouchers to ahead of the next event and assess village magnitude to assign the correct number of vouchers.

lukegb:
For the public export, we export:

* ID
* name
* URL (if any)
* location (if set)

For the private export, we also export:

* admin emails (same as for speakers/etc. for issuing vouchers)
* attendee count (from registration form)

We could use VillageMembership, but at the moment this will just be the admins because there's no way to join a village at the moment (so all village memberships should have is_admin=True, so the email will show up in the private export).

Fixes #1548

### DIFF
--- a/models/cfp.py
+++ b/models/cfp.py
@@ -24,6 +24,7 @@ from models import (
 from main import db
 from .user import User
 from .cfp_tag import ProposalTag
+from .village import Village
 from . import BaseModel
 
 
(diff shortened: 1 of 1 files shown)


Jonty:
Are we properly exporting village content with the link to the village elsewhere?

PR #1803, opened 2025-05-31
Fix specifying a URL when creating a village

lukegb:
Fix specifying a URL when creating a village


marksteward:
Partially fixed in #1851

PR #1807, opened 2025-07-20
Export lightning talk information as well

### LINKED ISSUE #1466: Lighting Talks are not correctly exported
Looks like lightning talks are not being correctly exported and are missing from the 2022 export: https://github.com/emfcamp/Website/blob/b00e10de33516730c5f6d8799f01d2ec179d929e/exports/2022/public/LightningTalkProposal.json

lukegb:
Lightning Talks do not go through the CfP proper, so they sit in the new state.

Hopefully the CfP redesign will get done before 2026, so this is mostly so we can export some semblance of sensible data from 2024.

Fixes #1466.

### DIFF
--- a/models/cfp.py
+++ b/models/cfp.py
@@ -9,7 +9,7 @@ from geoalchemy2.shape import to_shape
 from sqlalchemy.dialects.postgresql import ARRAY
 from sqlalchemy.ext.mutable import MutableList
 
-from sqlalchemy import UniqueConstraint, func, select, text
+from sqlalchemy import UniqueConstraint, func, select, text, or_
 from sqlalchemy.orm import column_property
 from slugify import slugify_unicode
 from models import (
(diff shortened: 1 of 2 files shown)


models/cfp.py
```diff
@@ -919,6 +924,33 @@ class LightningTalkProposal(Proposal):
     slide_link = db.Column(db.String, nullable=True)
     session = db.Column(db.String, default="fri")
 
+    should_be_exported = column_property(or_(
+        Proposal.is_accepted.expression,
+        Proposal.state == "new",
```
wlcx, line 929: Uninformed question: will this mean that any newly submitted other proposals would get exported or is that not what "new" actually means?

wlcx (approved):
Nothing looks weird but my review is not worth much :-)

PR #1813, opened 2025-08-12
cfp: Add needs-laptop and video privacy filters

### LINKED ISSUE #1693: Additional filter options in cfp-review
We need to filter by: 
- whether a laptop is required
- the selected recording status (ie. public, review, DNR)

wlcx:
image

Fixes #1693


SamLR:
rewriting the cfp system so lets hold off on this

wlcx:
I mean, it's not very complex, and if the cfp gets rewritten it just gets thrown out no? And if it doesn't then it's done

PR #1820, opened 2025-08-18
Switch to rsbuild

lukegb:
This substantially reduces the number of the JS dependencies we pull in
via package.json.

rsbuild (https://rsbuild.rs/) goes nyoom - this migrates the JS/TS
compilation from Babel to SWC, among other things, which helps
compilation go much faster.

To accommodate this, the "dist" directory is handed over to rsbuild, and
Flask is configured to serve static content from dist/static/ rather
than the root "static" directory, which is checked into Git. This is
because rsbuild is a bit smarter about trying to collect all the used
assets. Eventually it would be nice to stop copying _all_ of the assets
from e.g. octicons, and just copy the ones we need, but that's TBD.

The rsbuild config is actually smaller than our gulpfile. Huh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,13 +130,7 @@ node_modules
 .idea/
 
 # Compiled static assets
-static/*
 !static/financial-statements
 !static/fonts
 !static/deliveries.pdf
-
-# Hashed files created by flask-static-digest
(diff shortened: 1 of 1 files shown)


russss (approved):
Looks good to me. I think rsbuild is probably the right decision here.

I can't be sure this hasn't broken anything until we ship it, so let's do that...

Issue #1338, opened 2024-01-27
Generate apple/android wallet passes for tickets


wlcx:
Would be very cool. I've got a pass signing key from @eeeeeta (thanks eta, theta).

I've got basic generation working from json, so now it would be integrating it with the website.

My main question currently is how to handle the multiple tickets in a typical purchase receipt

eeeeeta:
> My main question currently is how to handle the multiple tickets in a typical purchase receipt

I heard you like zips, so I put some pkpass files (zips) in a bigger zip so you can zip while you zip:
https://developer.apple.com/documentation/walletpasses/distributing_and_updating_a_pass#3793284

wlcx:
Yeah I guess that works. Though, UX around having multiple admission tickets and not transferring them might suck a bit

lukegb:
I have access to Google Wallet pass generation on my Google account, but we should get it for an EMF merchant if we're going to do this.

castaway:
Not sure about Apple, but on Google Wallet you can add practically any QR / barcode, so does it need anything special?

wlcx:
Yeah looks like you can just issue them as-is on android: https://developers.google.com/wallet/retail/offers/overview/add-to-google-wallet-flow#creating-passes-objects-and-passes-classes-in-the-jwt

lukegb:
> Yeah looks like you can just issue them as-is on android: https://developers.google.com/wallet/retail/offers/overview/add-to-google-wallet-flow#creating-passes-objects-and-passes-classes-in-the-jwt

That JWT needs to be signed, and embedded in a URL.

will-h:
As discussed in https://github.com/pretix/pretix/issues/979 I think the Google Pay option has some privacy concerns, which isn't the case with simply generating a pkpass file offline. 

I use the 'PassAndroid' app for pkpass files on Android. Not a perfect solution.

wlcx:
> That JWT needs to be signed, and embedded in a URL.
So it does. 

> As discussed in https://github.com/pretix/pretix/issues/979 I think the Google Pay option has some privacy concerns, which isn't the case with simply generating a pkpass file offline.
mm, good point. Although - if we generate and sign the JWT ourselves (with keys google presumably issue us) I think that would avoid sending any user data to google? Until the user actually opens the link, of course

lukegb:
There's new-ish support for private passes, but this might only be supported for COVID/health related documentation, and the flow is quite annoying (doesn't sync between devices, etc.)

[...]

PR #1822, opened 2025-08-19
Add pkpass (apple wallet) entry pass generation

wlcx:
Generate pretty pkpass files with a user's entrance qr code. 

This shells out to openssl, but only with stuff in config as arguments so I don't think this is a problem.

Passes currently expire at the `EVENT_END` config option.

Resolves (at least the apple side) of #1338

Looks like this currently: 
image


wlcx:
[emf_ticket(52).pkpass.zip](https://github.com/user-attachments/files/21861073/emf_ticket.52.pkpass.zip)
(had to zip to appease github, despite the pkpass file itself just being a zip)

russss:
Checking in from the PR necromancy department:

I am in favour of this on the whole, but I don't like the number of assets with event branding which are specific to this feature. I'd be happy to merge it if there was some more generic branding.

russss (changes requested):
I should remember to do "request changes" just so it's clearer what's going on in the PR view...

russss:
This has been superseded by #2119

PR #1829, opened 2025-08-21
Fix up 2024 export so we don't consider everything as recorded

### LINKED ISSUE #1828: 2024 schedule appears to think everything was recorded
There's a video camera icon next to everything on https://www.emfcamp.org/schedule/2024

Oops.

This is because we always export the video attribute, but it usually contains `{"recording_lost": false}`.

lukegb:
This just deletes 'pointless' video attributes. There's a small change to the schedule generation, but that won't fix this alone - we should default the 'video recording lost' state in the database to NULL, so we can distinguish whether or not it's actually set to a useful value.

Fixes #1828.

This is annoying to review, sorry.


lukegb:
For review purposes: https://p.lukegb.com/NaturallyLoyalKangaroo.patch

(as generated by `diff -u <(jq . exports/2024/public/schedule.json) <(jj file show -r @-- exports/2024/public/schedule.json | jq '.')`)

PR #1830, opened 2025-08-21
Mix event date into SECRET_KEY on startup

### LINKED ISSUE #1827: User token reuse in future years
At the moment all [user tokens](https://github.com/emfcamp/Website/blob/main/models/user.py#L129-L166) (including checkin barcodes, bar training tokens, and user API tokens) are generated only with the user ID and our secret key.

Because the database is cleared every year, user IDs aren't preserved, and we don't change the secret key, this means a user can theoretically reuse their token to access aspects of different user's account in a future year. And also steal their tickets.

We could namespace all of these, but I wonder if the best approach is to somehow automatically incorporate the event year into the `SECRET_KEY`. I don't believe anything relies on it being stable between years, so we could just rotate it every year, but that adds room for error.

russss:
This ensures that the secret key changes between events, preventing old tokens from being reused.

Fixes #1827


lukegb:
This is a remarkably complicated way of doing `hmac.new(hmac.new(bytes([0]*32), b"sample_secret_key", hashlib.sha256).digest(), b"2024" + bytes([1]), hashlib.sha256).digest().hex()` :p (but I don't really object to this, using HKDF for this seems reasonable and probably avoids some other pitfalls).

I guess we should also (this time) rotate the key material once we land this just since otherwise we'll have previously used the HKDF key directly.

marksteward (approved):
lgtm

russss:
I would also note that we of course already have `cryptography` as a transitive dependency so there's no new library here, and it's worth using the right tool for the job I think.

PR #1833, opened 2025-08-22
Migration: replace deprecated/legacy Wise borderless account usage

jayaddison:
As far as I can tell, the notion of borderless accounts is deprecated in Wise's documentation; there are one or two trailing references to it, but most have been removed.

As a result, the `pywisetransfer` emits deprecation notices when those endpoints are used.  This pull request updates the website code to utilise replacement, non-deprecated endpoints.

Untested so far; opening as a draft.


jayaddison:
> Untested so far; opening as a draft.

I've begun testing this, and that process uncovered that the Wise API has some incompatible updates -- for example, in relation to the API responses from `balances` list operations that no longer contain bank address details directly.

Support for retrieving the relevant bank account info should be added in `pywisetransfer` by https://github.com/jayaddison/pywisetransfer/issues/38

russss (approved):

Issue #1831, opened 2025-08-21
Dependency audit


lukegb:
```
$ uv pip list --outdated

Package            Version Latest  Type
------------------ ------- ------- -----
apiron             8.0.2   9.0.0   wheel
cryptography       44.0.3  45.0.6  wheel
css-inline         0.14.6  0.17.0  wheel
email-validator    1.3.1   2.2.0   wheel
flask              2.3.3   3.1.2   wheel
flask-cors         5.0.1   6.0.1   wheel
flask-mailman      0.3.0   1.1.1   wheel
flask-migrate      3.1.0   4.1.0   wheel
flask-sqlalchemy   3.0.5   3.1.1   wheel
gunicorn           22.0.0  23.0.0  wheel
locust             2.38.1  2.39.0  wheel
lxml-stubs         0.3.1   0.5.1   wheel
pillow             10.4.0  11.3.0  wheel
pulp               2.1     3.2.2   wheel
python-memcached   1.59    1.62    wheel
python-stdnum      1.20    2.1     wheel
pyzmq              27.0.1  27.0.2  wheel
requests           2.32.4  2.32.5  wheel
ruff               0.12.9  0.12.10 wheel
sqlalchemy         1.4.54  2.0.43  wheel
stripe             8.0.0   12.4.0  wheel
unidecode          0.4.21  1.4.0   wheel
wtforms            2.3.3   3.2.1   wheel
wtforms-sqlalchemy 0.3     0.4.2   wheel
```

russss:
`flask-sqlalchemy` 3.1 depends on sqlalchemy 2. I'm inclined to also hang back on the `flask-migrate` upgrade until we've sorted that as well.

The situation is currently:

```
Package            Version Latest Type
------------------ ------- ------ -----
apiron             8.0.2   9.0.0  wheel
cryptography       44.0.3  45.0.6 wheel
flask-migrate      3.1.0   4.1.0  wheel
flask-sqlalchemy   3.0.5   3.1.1  wheel
pulp               2.1     3.2.2  wheel
sqlalchemy         1.4.54  2.0.43 wheel
wtforms            2.3.3   3.2.1  wheel
wtforms-sqlalchemy 0.3     0.4.2  wheel
```

`pulp` is a dependency of slotmachine which I will have a prod at sometime soon.

PR #1835, opened 2025-08-25
Dependencies: update pywisetransfer and transitive dependencies (apiron, cryptography)

jayaddison:
Progresses #1831.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -31,15 +31,15 @@ wheels = [
 
 [[package]]
 name = "apiron"
-version = "8.0.2"
+version = "9.0.0"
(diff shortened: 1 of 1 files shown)


russss (approved):

jayaddison:
If it's OK, I'd like to merge this and #1833 despite the fact that the results are not yet completely functional.  To explain: while testing that latter PR I learned that the Wise API has changed since the Y2024 event, and adapting to those changes will require some maintenance work here and some feature work in `pywisetransfer`; particularly: https://github.com/jayaddison/pywisetransfer/issues/38

PR #1843, opened 2025-08-26
Remove server default

marksteward:
These are all because it's easier to add a column default than write a migration, but it's safe to fix now the DB's empty.


marksteward:
Depends on #1842, oops.

PR #1847, opened 2025-08-28
payments: update Wise implementation to match API changes

### LINKED ISSUE #1844: Wise API breakage
> If it's OK, I'd like to merge this and #1833 despite the fact that the results are not yet completely functional.  To explain: while testing that latter PR I learned that the Wise API has changed since the Y2024 event, and adapting to those changes will require some maintenance work here and some feature work in `pywisetransfer`; particularly: https://github.com/jayaddison/pywisetransfer/issues/38

_Originally posted by @jayaddison in https://github.com/emfcamp/Website/issues/1835#issuecomment-3222881555_

jayaddison:
As described in #1844, the Wise API has been updated and we need to update our codebase accordingly.

An open question is whether our `BankAccount` records for Wise accounts should use the same account-per-currency design model as our Stripe accounts.  Currently, they still do, but I would like to explore whether it would make sense to consolidate to a single account ID (on which we could, for example, populate all of sort-code, account-number, SWIFT and IBAN).

Creating this pull request as a draft, initially.

Resolves #1844.

### DIFF
--- a/templates/about/contact.html
+++ b/templates/about/contact.html
@@ -8,7 +8,7 @@
     <li><a href="mailto:content@emfcamp.org">Content</a>, for talks, workshops, and installations</li>
     <li><a href="mailto:volunteers@emfcamp.org">Volunteering</a></li>
     <li><a href="mailto:sponsors@emfcamp.org">Sponsorship</a></li>
-    <li><a href="mailto:vendors@emfcamp.org">Food/drink vendors</a></li>
+    <li><a href="{{url_for('base.page', page_name='traders')}}">Traders</a></li>
     <li><a href="mailto:villages@emfcamp.org">Villages</a></li>
   </ul>
 
--- a/templates/about/food.md
+++ b/templates/about/food.md
@@ -7,7 +7,8 @@ Don’t forget to eat and drink - hunger and dehydration are the enemies of havi
 We have a great range of food vendors with vegetarian, vegan, and gluten free options. Doing a [volunteer shift](/volunteer/info) will allow you to enjoy a meal cooked by our excellent crew catering!
 
 ### Vendors
-
+A full list of food vendors will be available closer to the event.
+<!--
 * [Cantina El Burrito](https://www.cantina-elburrito.co.uk/) - Tasty Mexican burritos, nachos and chilli.
 * [Colombo Street](https://www.colombostreet.co.uk/foodtruck) - Award Winning Sri Lankan street food.
 * [Goodness Gracious Healthy Foods](https://www.mobileveggiefood.co.uk/) - Fresh, healthy vegan and vegetarian food, with plenty of falafel, halloumi and hummus.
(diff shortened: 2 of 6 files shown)


apps/payments/wise.py
```diff
@@ -203,53 +203,57 @@ def wise_business_profile():
...
-            continue
 
-        address = ", ".join(
-            [
-                account.bankDetails.bankAddress.addressFirstLine,
-                account.bankDetails.bankAddress.city + " " + (account.bankDetails.bankAddress.postCode or ""),
-                account.bankDetails.bankAddress.country,
-            ]
-        )
+            bank_name, _, bank_address = bank_info.partition("\n")
```
russss, line 235: This feels like it might come back to bite us at some point - perhaps we should remove the institution column from BankAccount and just throw this all in the address field.

apps/payments/wise.py
```diff
@@ -203,53 +203,57 @@ def wise_business_profile():
     return id
 
 
+def _retrieve_detail(details, requested_type):
```
russss, line 206: Is there a reason why we can't make pywisetransfer return this as a dict (or even a dataclass) for us? Not sure what the API response looks like.

apps/payments/wise.py
jayaddison: Yep, this does seem fragile.  However: it does seem useful in the admin UI (and perhaps elsewhere?) to have a small descriptive text field with the institution name alone.

I'll think about it some more.  Another option could be to add validation logic so that we detect any likely breakage early.

apps/payments/wise.py
jayaddison: (note also: that was gathered from the Wise Sandbox env)

apps/payments/wise.py
```diff
@@ -203,57 +208,102 @@ def wise_business_profile():
...
-                account.bankDetails.bankAddress.country,
-            ]
-        )
+        bank_details = _combine_account_recipient_details(account)
+        if not bank_details.name_and_address:
+            # TODO: add an error or warning about this condition
+            continue
 
-        sort_code = account_number = None
+        bank_name, _, bank_address = bank_details.name_and_address.partition("\n")
```
jayaddison, line 279: TODO: relocate this to a `@property` on the recipient dataclass(es).

apps/payments/wise.py
```diff
@@ -203,57 +208,102 @@ def wise_business_profile():
...
-        address = ", ".join(
-            [
-                account.bankDetails.bankAddress.addressFirstLine,
-                account.bankDetails.bankAddress.city + " " + (account.bankDetails.bankAddress.postCode or ""),
-                account.bankDetails.bankAddress.country,
-            ]
-        )
+        bank_details = _combine_account_recipient_details(account)
+        if not bank_details.name_and_address:
+            # TODO: add an error or warning about this condition
```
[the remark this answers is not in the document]
jayaddison: Resolved by adding inline assertions instead.

russss (approved):

jayaddison:
Thanks @russss!

PR #1848, opened 2025-08-28
Update travel.md

grajohnt:
Modestly expanded info from the perspective of those less familiar with the UK and EMF.

### DIFF
--- a/templates/about/travel.md
+++ b/templates/about/travel.md
@@ -4,17 +4,21 @@ title: Travelling to Electromagnetic Field
 
 EMF will be held at the Eastnor Castle Deer Park, just east of Ledbury, Herefordshire.
 
+To experience the full event, you should aim to arrive on Thursday 16 July and depart mid-day Sunday 19 July, with presentations beginning on Friday 17 July. You won't be allowed on site outside of those times unless you've [volunteered](/about/volunteering) to help with setup or teardown.
+
 {# 2024
 You can arrive after 10am on Thursday 30 May and you can stay until midday on Monday 3 June. You won't be allowed on site outside those times unless you've [volunteered](/about/volunteering) to help with setup or teardown.
 #}
 
 ## By Train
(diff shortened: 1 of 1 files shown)


grajohnt:
> This should probably say "depart mid-day on Monday". We have content running until late on Sunday every year.

Modified to reflect this. Date range for presentations is also included because time slots are often not announced until late. When planning travel, it's good to know that as long as you arrive on Thursday anytime and leave Monday, you won't inadvertantly miss it.

templates/about/travel.md
```diff
@@ -4,17 +4,21 @@ title: Travelling to Electromagnetic Field
...
+If you are unfamiliar with the UK train system, please refer to [this wikivoyage page](https://en.wikivoyage.org/wiki/Rail_travel_in_Great_Britain) and follow their general guidance.
+
+EMF is close (5 km) to Ledbury station which is around [three hours](https://traintimes.org.uk/london+paddington/ledbury/) from London Paddington (on [GWR](https://www.gwr.com)) and [one hour](https://traintimes.org.uk/birmingham+new+street/ledbury) from Birmingham New Street (on [West Midlands Railway](https://www.westmidlandsrailway.co.uk/)).
 
 The station at Ledbury has short platforms, so if you are on one of the long 9-coach trains direct
 from Paddington, you will need to be in the front three coaches to get off the train.
 
-We will be providing a free shuttle bus service between Ledbury station and the EMF site.
+We will be providing a free shuttle bus service between Ledbury station and the EMF site during the event.
```
jayaddison, line 21: Nitpicky thought/suggestion: how about saying "throughout the event" instead of "during"?  I feel that would make it clearer that it is an ongoing/continuous service as opposed to a one-off bus.

(the following paragraph about the timetables does already imply that fairly clearly though, I admit)

grajohnt: The motivation for these edits was to provide as much clarity as possible to those who must book travel far in advance, and so being as specific as possible is good. The shuttle bus schedule doesn't get announced until late, so I agree that this isn't as clear as it could be.

Perhaps it should be something like:
...providing free and regular (or frequent?) shuttle bus service...during the event.

This might help people understand that you're not stuck at Eastnor - it's possible to buy food etc in Ledbury if needed.

jayaddison: No worries, that makes sense.  The `free and regular` phrasing has a more pleasant tone to it, I think (something about the duplicate f-sound reads oddly -- even if it would be maybe marginally less ambiguous)

PR #1849, opened 2025-08-28
Fully typed models

wlcx:
These changes make full use of SQLAlchemy 2's typing support and greatly improve our type checking situation.

The main bits:
- Replace model column definitions with `mapped_column`, with explicit python types.
- Replace accesses to sqlalchemy via flask_sqlalchemy's `db` with direct imports from sqlalchemy itself.
- Replace uses of `backref` with `backpopulates` and explicit `relationship` on both side of the relation. This enables full typechecking from both sides.
- Replace all the queries using the old style `Model.query...` (can't be type checked) with `select(Model)...`
  - There's plenty of the former still left, just not in places mypy is currently checking. These can be more gradually migrated as we improve type coverage.

And supporting changes to fix type errors/improve type checking:
- Make some nullable things that shouldn't be nullable non-nullable
- Make ShiftEntryState and Currency enums
- Split admin form venue/village locations into separate lat/long float fields
- Sprinkly type narrowing, casts, and type ignores (in order of preference) to fix type errors

This is a big PR - I'd be very much open to moving the enum stuff (for instance) into separate PRs.
In terms of risky bits, IMO the most potential for introducing bugs is in the query changes.


models/scheduled_task.py
```diff
@@ -46,11 +48,11 @@ def __repr__(self):
...
-    id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String, nullable=False)
-    start_time = db.Column(db.DateTime(True), nullable=False)
-    duration = db.Column(db.Interval, nullable=False)
-    result = db.Column(db.JSON, nullable=False)
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column()
+    start_time: Mapped[datetime] = mapped_column(DateTime(True))
+    duration: Mapped[timedelta] = mapped_column(Interval)
+    result: Mapped[dict[str, Any]] = mapped_column(JSON)
```
marksteward, line 55: Does `Mapped[JSON]` work here?

wlcx: I mean it doesn't cause an error, but I think any uses of `result` are in untyped functions atm, so anything works here. AFAICT `dict[str, Any]` is what this ends up being though?

wlcx: Or are you meaning that it could technically not be a dict and should be typed as any json value?

Can you even do soemthing like 
```python
type JSONValue = None | str | int | float | list[JSONValue] | dict[str, JSONValue]
```

:p

marksteward: Yeah, but I think we should probably be using a custom JSON type that supports mutation anyway, so I'll add that to my list.

wlcx: yeah, even better

Issue #1778, opened 2024-09-08
Fix code scanning alert - URL redirection from remote source


marksteward:
Use `get_next_url()` instead.

Tracking issue for:
- [ ] https://github.com/emfcamp/Website/security/code-scanning/50

PR #1857, opened 2025-09-13
Clean up next URL handling

lukegb:
Fixes https://github.com/emfcamp/Website/security/code-scanning/50 and #1778

### DIFF
--- a/apps/users/__init__.py
+++ b/apps/users/__init__.py
@@ -1,6 +1,4 @@
 import time
-from typing import overload
-from urllib.parse import urljoin, urlparse, urlunparse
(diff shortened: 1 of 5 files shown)


apps/users/__init__.py
```diff
@@ -59,17 +60,27 @@ def users_variables():
     }
 
 
-def is_safe_url(target):
+def make_safe_url(target: str) -> str | None:
```
wlcx, line 63: worth moving to apps.common maybe?

wlcx (approved):
Seems legit

PR #1863, opened 2025-10-20
Village long description

### LINKED ISSUE #1816: Village details page
Give villages a bit more (non-wiki) space with which to express themselves (and maybe finish the membership feature?)

Probably some sort of markdown description, to go along with the short summary.

ashirley:
**There is still a lot to tidy up here but wanted to get some opinions sooner rather than later about iFrame layout options**

Add a long_description field to villages. This is markdown from untrusted users and is rendered using nh3 and a sandboxed iFrame in order to limit the untrusted HTML.

There are 2 layout options for the iFrame:

The first uses javascript to resize the iFrame to prevent scrolling within the iFrame. 

vivaldi_2gFsXcJgpn

The second uses a static size for the iFrame which scrolls for long content.

vivaldi_UirExic3aR

The advantage of the second is that we can remove the allow-scripts privilege from the iFrame.

fixes #1816


wlcx:
> unexpected status from HEAD request to https://registry-1.docker.io/v2/library/node/manifests/12-alpine: 500 Internal Server Error

us-east-1 detected

PR #1865, opened 2025-11-26
Remove broken sponsor purchase links

Mattsi-Jansky:
Right now these links aren't working because tickets aren't on sale. When tickets are available in the future, this can be reverted.


russss (approved):
I'll just merge this for the moment and maybe we can make it more fancy when we re-add it.

PR #1881, opened 2026-03-03
Tests: payments: dynamic profile_id retrieval

jayaddison:
The hardcoded `profile_id=123456789` previously in place here is in conflict with the ability to regenerate pytest-vcr fixtures.

Regeneration of those fixtures requires a valid Wise sandbox token, but the sandbox environment does also restrict the profile IDs accessible by those tokens -- so we need to retrieve information for a valid account linked to the token.

To do that, retrieve the (business) profile ID dynamically.

Follow-up related to #1847.


jayaddison:
Superseded by #1883.

PR #1940, opened 2026-03-27
 Add support for admonition and toc extensions to mark down.

unknowndomain:


russss:
This will need squashing before merging because the merge queue doesn't give us that option.

PR #1942, opened 2026-03-28
Tweaks for development and review environment

lukegb:
* Make development envs automatically populate fake data (opt-out)
* Enable `BYPASS_LOGIN` in review environments
* Allow "@test.invalid" emails in review environments as well (since their mail goes to MailHog anyway)


marksteward (approved):
Thank you, I've been meaning to look at this for ages

PR #1951, opened 2026-04-05
Port the buildup safety manual to Markdown

lukegb:
This will change anyway before we actually start using it, but I figure
it's easier to edit this prose in Markdown rather than in HTML.

This uses a little Jinja helper to automatically markdownify anything
that's included from a Jinja template where the filename ends in .md
before it goes through the templating engine.


marksteward:
The admin email preview code also currently does Jinja-then-markdown, and I'm beginning to think it's the only way it can work.

Markdown will wrap every line in paragraphs or `<br/>`, and (in the email case) apply smart quotes. This applies even within Jinja tags, which you want to be left untouched. On the other hand, Markdown explicitly allows HTML input (in the right context) so calling Jinja first is safe.

marksteward:
For once I'm not worried about security! By safe, I mean from getting ourselves into an awkward position where we have to escape input or restructure it to work around annoying/confusing behaviour. Markdown-then-Jinja works for static files, so maybe we just drop a warning if there are any Jinja tags inside the included file.

Or we could use a custom block like `{% include_markdown ... %}`?

russss (approved):

PR #1955, opened 2026-04-06
Site wide admonitions

unknowndomain:
1. Removes legacy `.alert` and `.alert-{type}`classes.
2. Moves .admonitions back to base.scss
3. Adds .alert classes to the admonition ones so they can be used interchangably.
4. Moves the various example of `border-radius` into a variable for consistency as they were between 3-6px across the site.
5. Moves various uses of `.well` to `.alert`.


russss:
Thanks! Can you please provide screenshots of all the pages you've changed, and also squash your changes into a single commit, as we're not able to squash commits on merge.

russss (approved):

russss:
Thanks. I've made a few changes to how the variables work and changed the colours around so they're not too bright.

unknowndomain:
The issue with the new colours is the contrast with the icons.

There should be examples here: http://localhost:2342/dev/design/markdown

Screenshot 2026-04-12 at 18 16 45@2x

Screenshot 2026-04-12 at 18 17 13@2x

Screenshot 2026-04-12 at 18 17 44@2x

Screenshot 2026-04-12 at 18 18 04@2x


Hopefully we can fix this with the space theme update once we have the design for that?

russss:
Yeah, I spotted that and decided it was good enough for the moment. We'll hopefully have a new theme soon.

PR #1961, opened 2026-04-09
Role and venue slugs

jellybob:
This should be reasonably inoffensive, it updates things like bar training to use stable slugs for roles and the role config to use slugs for venues. The role migration deletes any volunteer role/shift data because I foolishly changed seed data at the same time, and given we've not got any of this active in production yet I couldn't bring myself to write a clean migration, if anyone is sufficiently upset by that I'll give it a shot. Also added a few cascades in appropriate places to support deleting roles cleanly.


marksteward:
Is it worth using hyphens so we can use them in URLs?

russss (approved):

PR #1965, opened 2026-04-11
Allow longer village names

### LINKED ISSUE #1962: Allow longer village names
Hi, it would be nice if the max length on village names could be increased? It's currently at 25 characters which is slightly too short for the people of Glastonledburyshire-on-Severn. Many thanks.

edwinbalani:
Fixes #1962.

I picked the length of a (classic) tweet arbitrarily.

The [Village model](https://github.com/emfcamp/Website/blob/52701ba379d3bdc1007a58ca934dc9581d4c8bdf/models/village.py#L32) doesn't constrain name length, so just the form needs changing.

### DIFF
--- a/apps/villages/forms.py
+++ b/apps/villages/forms.py
@@ -17,7 +17,7 @@ from ..common.forms import Form
 
 
 class VillageForm(Form):
-    name = StringField("Village Name", [Length(2, 25)])
+    name = StringField("Village Name", [Length(2, 50)])
     description = TextAreaField("Description", [Optional()])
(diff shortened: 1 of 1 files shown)


russss:
I think 140 characters is too long because we want to show these on the map and it'll cause pain. 50 characters sounds like a decent compromise.

edwinbalani:
Sounds good to me, I did pick a silly number to elicit feedback.

If the concept doesn't fit in 50 characters people'll have to use logograms or emoji or something

russss (approved):

PR #1966, opened 2026-04-12
Account page tidy

jellybob:
Apply the styling built for the [supprting EMF](https://emfcamp.org/supporting-emf) page to the main account page:

- Panels in the grid are the same height for each panel in the
  same row.
- Actions exist in a panel footer, which makes panels with
  significantly differing heights look a bit more balanced.
- Replace action buttons with links.

I've not touched the wider page layout as that's being worked on in #1963 

Screenshot 2026-04-12 at 14 44 07

### DIFF
--- a/css/_panel_grid.scss
+++ b/css/_panel_grid.scss
@@ -1,4 +1,5 @@
 /* A grid of panels. */
+@use "./_variables.scss" as *;
 
 .panel-grid {
   --panel-grid-min-size: 350px;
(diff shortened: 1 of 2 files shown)


russss:
Sorry, but I don't like this style at all - the links aren't obvious enough. I let this slide for the Supporting EMF page because it's a standalone page and we don't have time for a full style refresh this year, but I want to keep the buttons on the rest of the site for this year.

jellybob:
That's fair. How do you feel about leaving buttons, but having them in the footer of equal sized panes? I don't feel strongly on links vs buttons but I cringe a little any time I see that page and its a hodgepodge of randomly sized panes.

russss:
I'm happy with the equally-sized panels - I was going to do that anyway after #1963 was merged.

jellybob:
Now with buttons.

Screenshot 2026-04-12 at 14 59 49

russss:
Apparently I can't re-open this so you'll have to create a new PR? Sorry about that

PR #1977, opened 2026-04-22
Initial wiki implementation

futureshape:
Implements #1949 - leaving it as draft as there are several architectural changes yet to be decided, and I still need to test some paths, but though I'd ask or early feedback.

Created a new `/wiki` app to serve wiki pages. Proposing to keep that flexibility rather than assume we put all wiki pages in `/about` or similar.

# Main UI changes

## Wiki landing page

`/wiki` shows a list of pages and an option to create a new page, which is restricted to admins

image

## Create page

`/wiki/new` creates a new page - I think it's better to do this through a form than to require a deployment for new wiki pages. Slug and title can be set independently. 

image

## View page

`/wiki/<slug>` renders the page using the same markdown to HTML method as the village pages. Option to edit restricted to logged in members.

❓ Do we want to restrict this to only logged in people who *have a ticket*?
❓ Do we want to display editor names here? (also applies to edit history)

image

## Edit page

`/wiki/<slug>/edit` lets people edit the page
 
 ❓ Should we allow page title changes, and if yes maybe restrict this to admins only?
 ❓ Do we actually need edit summaries?
  
image

## View page history

`/wiki/<slug>/history` shows the page history and links to comparison (diffs)

image

## Compare page versions

`/wiki/<slug>/diff/<v1>/<v2>` shows a markdown diff in the standard format

image

# Other changes

Moved `render_markdown` (untrusted markdown renderer) to `apps/common/__init__.py` together with other markdown functionality so it can be used by both villages and wiki. 

# TODO

- [x] Test edit conflict functionality (implemented but not yet tried it on local dev)
- [ ] Add tests

Assisted-by: Claude Sonnet 4.6


apps/common/__init__.py
```diff
@@ -449,6 +451,23 @@ def get_next_url(default=None):
     return default
 
 
+def render_markdown(markdown_text: str) -> Markup:
```
marksteward: Could this go next to `render_template_markdown`, and perhaps be renamed `render_untrusted_markdown`?

marksteward (changes requested):
GitHub seems to be struggling atm

marksteward:
I think a force feature is risky. We should just update the token to the current one having applied any changes and let them try again. If it conflicts again they probably want to know instead of stomping over it.

marksteward:
`created_at`, `updated_at`, `created_by_id`, and `created_by_id` are all available from the version/transaction table. I'm not a fan of duplicating them into every version, especially as they're not used currently.

marksteward:
Please add the `import wiki` in main.py to the bottom of `models/__init__.py` instead

futureshape:
Updates based on feedback above and discussion on IRC:

* Code style updates
* Removed link to "All wiki pages" since it's not a useful index for most users (still accessible manually on /wiki)
* Editor names are only shown to admin users
* Removed redundant timestamp fields and edit summary
* Improved handling of edit conflicts using https://pypi.org/project/merge3/
* Only admins can change page titles

marksteward:
Rebased and merged with some minor fixes as 7c991295

marksteward:
This is really great, thanks!

PR #1978, opened 2026-04-24
1526 village members

### LINKED ISSUE #845: Allow people to add themselves to villages
A few outstanding questions here:

* Why? What advantages does this give people?
* How much info can village owners/other village members/the general public see on village members?

Ref #667

### LINKED ISSUE #998: Add "leave village" functionality
People will make the wrong village by mistake, or might want to change name. If everyone leaves a village, it should be deleted.

### LINKED ISSUE #1526: Allow village administrators to add other people as village administrators
We need village administrators to be able to add other people as village administrators so they can add/edit the village scheduled content. This is causing a problem in larger villages with their own content (like milliways) as it means one person is a bottleneck.

This isn't quite the same thing as people adding themselves to villages, which is https://github.com/emfcamp/Website/issues/845

ashirley:
Orga can manage non-admin members of a village, users can join a village themselves, a village-admin can promote someone who has joined their village to an admin.

This should add-up to attendees being able to self-serve giving each other access to edit the village's long_description (and other fields though these are less likely to need frequent editing)

Are the following things OK?
* ~~There is no way to remove a village admin apart from via orga~~
  * admins can now leave
* ~~There is no way to leave a village you have joined apart from via orga (and you can't join multiple villages) (see #998)~~
  * All members can now leave
* When you join a village the admins can see your name and email, ~~this is communicated via a tooltip only (but is probably to be expected?)~~
  * There is a confirmation page which makes this clear now
* This only lets village-admins schedule content in village venues (but includes a column I don't understand but doesn't seem sensitive for anyone in a village)
* Village admins can promote members to be admins without their consent. I can't see a problem in most scenarios but when the last admin tries to leave a village they have to promote someone first. That could make someone the sole admin of a village they didn't create and give them the unwanted responsibility for it (or for choosing to delete it).

Screenshot_20260424_085509
Screenshot_20260504_232704
Screenshot_20260504_232852
Screenshot_20260424_085604

Screenshot_20260504_233605

Screenshot_20260504_233811

Screenshot_20260425_213617
Screenshot_20260425_213634
Screenshot_20260425_214909

Fixes #1526
Fixes #845
Fixes #998


futureshape:
> When you join a village the admins can see your name and email, this is communicated via a tooltip only (but is probably to be expected?)

From a usability point of view, I would change it to something more explicit, maybe just a native `window.confrim`. Tooltips may not show reliably in touch devices.

> There is no way to leave a village you have joined apart from via orga

I would include this here, because it’s inevitable that someone is going to press the Join button not understanding what they’re doing, and they would want to leave. Fine to only join one village at a time though, it’s a bit of a physical constraint if the idea is that you camp with a village.

ashirley:
> > When you join a village the admins can see your name and email, this is communicated via a tooltip only (but is probably to be expected?)
> 
> From a usability point of view, I would change it to something more explicit, maybe just a native `window.confrim`. Tooltips may not show reliably in touch devices.

Good point about mobile users. In hindsight it is trivial to add a confirmation screen so I've done that.

> > There is no way to leave a village you have joined apart from via orga
> 
> I would include this here, because it’s inevitable that someone is going to press the Join button not understanding what they’re doing, and they would want to leave. Fine to only join one village at a time though, it’s a bit of a physical constraint if the idea is that you camp with a village.

Added now (including deleting a village when the last person leaves and preventing the last of the admins leaving without promoting a normal member first)

futureshape:
Thanks! from a quick scan of the code, it looks like all paths that I can think of are cover now :)

russss:
Thanks for doing this!

I know it doesn't make much practical difference, but I think we do need an approval step before someone can join a village. There are plenty of villages which aren't really open membership, and I think it will give people the wrong impression if they can join them without approval.

I think the best way of doing this is to create a `VillageJoinRequest` table, change the button to "request to join" and then have an admin page for village admins to approve these requests. We can probably do without email notifications for the moment.

ashirley:
Makes sense. Do you think villages should be able to choose to be open-membership if they want (to cut down admin and encourage people joining)?

Also, I assume you can only request membership of one village at a time (and can't request to join if you're already a member of another etc).

russss:
> Do you think villages should be able to choose to be open-membership if they want (to cut down admin and encourage people joining)?

I think if the button says "Request to join" then that will probably be enough discouragement for the moment. I don't want to add more complexity than necessary, and we can always add this feature later.

> Also, I assume you can only request membership of one village at a time (and can't request to join if you're already a member of another etc).

Yeah, we're still saying one-village-per-person for the moment. But I guess that means that you also need to be able to withdraw a join request.

marksteward:
Looks good! There's some minor clunkiness with the buttons for long village names:

image

And I think we should prevent deletion of a village with content, just as a precaution, but I think attendee content is currently broken.

marksteward (approved):

PR #1982, opened 2026-04-27
When accepting proposals that grant tickets, issue a voucher instead

### LINKED ISSUE #1971: CfP purchase changes
We don't want people to abuse the CfP guaranteed tickets.

* When a user's first Proposal is accepted, we should record the total number of adult tickets held by the proposer (`n`) against the user
* The "CfP-only" ProductView should only allow the proposer to purchase `(2-n)` adult tickets.
* The CfP acceptance email should inform the proposer of that fact

lukegb:
This allows us to fix #1971 and #1957: #1957 is fixed by only issuing a voucher if the proposal type tells us to, and #1971 is fixed because we have a lot of logic already around how to handle vouchers properly, and we don't need to separately do accounting for this.


SamLR (approved):
Looks good I'm going to raise a PR updating the email text for the case where someone should be given a voucher but isn't as they already have > 2 tickets

models/content/cfp.py
```diff
@@ -390,6 +391,26 @@ def mark_messages_read(self, user):
...
+
+        if self.type_info.grants_event_tickets and self.user.cfp_voucher is None:
+            # Issue a pseudo-voucher, which may have 0 capacity if the user already has 2 tickets.
+            product_view = ProductView.get_by_name("speakers")
+            if not product_view:
+                raise Exception("No 'speakers' product view created yet?")
+            voucher = Voucher(
+                view=product_view,
+                email=self.user.email,
+                tickets_remaining=max(2 - self.user.admission_tickets_held, 0),
```
russss: This should only count adult tickets owned by the user. (And specifically the `Product.is_adult_ticket(voucher=True)` definition of "adult".)

russss (changes requested):

russss (approved):

PR #1989, opened 2026-05-02
Add round object to track CfP round info

### LINKED ISSUE #1097: track round ranks and closing counts
we should track what vote 'value' we use as the cut off for each round

mostly because we forget and this is useful information round to round

SamLR:
Adds the 'Round' and 'ProposalRound' tables to track 

* min number of votes
* min score
* type of round close (talk or workshop)
* per proposal
    * number of votes
    * score
    * outcome (not enough votes, accepted, unaccepted, rejected and in-progress for between vote counting & ranking)

**NB** Currently these objects only exist in the database and as part of the round-close->rank flow. There's another issue, #1990, to actually make these usable in subsequent rounds

Closes #1097


models/content/cfp.py
```diff
@@ -124,6 +171,32 @@ class ReviewType(StrEnum):
 ]
 
 
+class ProposalRound(BaseModel):
+    __tablename__ = "proposal_round"
+    proposal_id: Mapped[int] = mapped_column(ForeignKey("proposal.id"), primary_key=True)
+    round_id: Mapped[int] = mapped_column(ForeignKey("round.id"), primary_key=True)
+    votes: Mapped[int]
```
marksteward: Can this be `vote_count`? `votes` would be a collection of ProposalVote instances

models/content/round.py
```diff
@@ -0,0 +1,35 @@
...
+from .cfp import ProposalRound
+
+
+class Round(BaseModel):
+    __tablename__ = "round"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+
+    created: Mapped[datetime] = mapped_column(default=naive_utcnow)
+    # Used for sorting in the CfP review page
```
marksteward: Is this true?

models/content/round.py
```diff
@@ -0,0 +1,35 @@
...
+    modified: Mapped[datetime] = mapped_column(default=naive_utcnow, onupdate=naive_utcnow)
+
+    proposal_type: Mapped[str]
```
marksteward: Should this be ProposalType?

apps/cfp_review/base.py
```diff
@@ -1195,9 +1197,9 @@ def schedule_item_change_owner(schedule_item_id: int) -> ResponseReturnValue:
     )
 
 
-@cfp_review.route("/close-round", methods=["GET", "POST"])
+@cfp_review.route("/close-round/<proposal_type>", methods=["GET", "POST"])
 @admin_required
-def close_round():
+def close_round(proposal_type: ProposalType):
```
marksteward: Please validate ProposalType early on if it's typed like this.

apps/cfp_review/base.py
```diff
@@ -1220,14 +1222,27 @@ def close_round():
     if form.validate_on_submit():
         if form.confirm.data:
             min_votes = session["min_votes"]
+            round = Round(proposal_type, int(min_votes))
+            db.session.add(round)
+            db.session.commit()
```
marksteward: Why the early commit?

apps/cfp_review/base.py
```diff
@@ -1260,20 +1275,17 @@ def close_round():
 @cfp_review.route("/rank", methods=["GET", "POST"])
 @admin_required
 def rank() -> ResponseReturnValue:
-    proposals_query = select(Proposal).where(Proposal.state == "reviewed")
+    round = get_or_404(db, Round, session["round_id"])
```
marksteward: This is a great way to get stuck in dev mode. Better to clear round_id if it's invalid (and handle the empty value properly).

Also this forces the session visiting /rank to be the session closing the round. Is it clearer just to pass as an argument to the handler?

marksteward (changes requested):

apps/cfp_review/base.py
SamLR: because we want the `round_id` on line 1234

models/content/cfp.py
```diff
@@ -124,6 +171,32 @@ class ReviewType(StrEnum):
...
+    score: Mapped[float]
+    outcome: Mapped[str]
+
+    round: Mapped["Round"] = relationship(back_populates="proposals")  # noqa UP037
+    proposal: Mapped["Proposal"] = relationship(back_populates="rounds")  # noqa UP037
+
+    def __init__(
```
marksteward: Please don't create overrides that just fill in every field. SQLa provides default initialisers that are more flexible.

apps/cfp_review/base.py
marksteward: You can just set `.round` on the ProposalRound, and even pass it into the initialiser if you haven't created an override that hides it.

marksteward (changes requested):

models/content/round.py
SamLR: removed

russss (approved):

PR #1999, opened 2026-05-03
Send CfP emails via a queue

### LINKED ISSUE #1997: Use email queue to send CfP round-close emails
At the moment they're sent in-line and this is quite slow.

marksteward:
We often send a bunch of emails for CfP acceptance, which clogs up admin processes and can fail. Inserting them into a DB table also means we get the same transaction lifetime, so we don't need to worry about periodic commits/correctness.

Emails don't get sent until the periodic task is run (`./flask periodic`), so we should probably run that inside docker-compose. They can also be delayed up to a minute as a result, so it might be worth adding a way of poking the scheduler.

There's currently no process to delete sent emails, which we'd probably want to be determined by the caller.

Fixes #1997.


russss:
Is this ready for review? It would be good to get this merged because I think we can use it to send the finalisation emails from the web interface.

russss (approved):

PR #2018, opened 2026-05-17
2026 branding

lukegb:
Let's go to space


Kunsi:
Should we update /about/branding at the same time?

lukegb:
Form well: 
image

Shift list (colourless): 
image

Shift list (colourful - manually forced just for demo): 
image

Volunteer admin: 
image

PR #2023, opened 2026-05-22
add fridge/freezer info to accessibility page

jchoules:
- [x] Check that Info lead is happy with wording before merge


marksteward:
Wording checked. I wonder if it's worth linking to this from the Childcare/Health sections

PR #2026, opened 2026-05-23
Fix Occurence fakes

jellybob:
They no longer have a state field which was causing starting from a blank database to fail.


russss:
Fixed in 3eff558f which was committed but not pushed, sorry about that!

PR #2030, opened 2026-05-24
Volunteer schedule view improvements

jellybob:
Improvements to the volunteer schedule view:

- [x] Replace colourful mode with icons indicating whether a shift is sufficiently staffed.
- [x] Use row colour to indicate clashes with an attendee's starred content or other shifts.
- [x] Add a legend for what icons/colours are indicating.
- [x] Make the role list selector less overwhelming.

To be done on future branches:
- [ ] Add something to the main logged in user page indicating upcoming shifts.
- [x] Hide shifts that conflict with shifts the user is already signed up for.


marksteward:
This looks amazing! Still looking at it, but clicking on a shift seems to suggest it's a conflict with itself (and uses 12-hour times instead of 24-hour)?

image

marksteward (approved):

jellybob:
> This looks amazing! Still looking at it, but clicking on a shift seems to suggest it's a conflict with itself (and uses 12-hour times instead of 24-hour)?

I guess strictly it does but also we probably shouldn’t pop up and say that! I’ll go shout at JavaScript time formatting a bit as well.

PR #2031, opened 2026-05-24
Closing date for CFP

katiesteckles:
We're not sure what {# 2026 #} is doing so please do something with that if needed


russss:
The "{# 2026 #}" is so that we find that date when we search the codebase for "2026" before the next event 🙃

PR #2034, opened 2026-05-28
Add 2026 branding info ahead of the site redesign

ashirley:
This is as a result of a meeting of the design team. There is no more information than we supplied for 2024 (apart from the caveat about gradient accessibility which didn't apply then).

Screenshot_20260601_223916

The relates to #2018 which is about changing the branding of the site itself.

~~Draft while I confirm a few things with the wider design team~~


templates/about/branding.html
```diff
@@ -52,91 +52,107 @@ <h3>Sizes</h3>
   The following styling guidelines refer to the 2024 event. Our new, <a href="https://blog.emfcamp.org/2025/11/28/theme-announcement/">space-inspired</a> theme for 2026 will be published soon.
```
marksteward: Should this be removed/reworded?

marksteward (changes requested):

marksteward:
🚀

PR #2047, opened 2026-05-31
Add link to blog post

willselby:
We need to add wording to the accepted e-mail to direct speakers to the relevant blog post.


willselby:
@katiesteckles @mscroggs

russss:
I'm a bit reticent to add year-specific links to emails, as they can often get missed at the next event.

What problem is this trying to solve?

katiesteckles:
Anyone who books now (which won't be many but will be anyone who gets a speaker ticket) will have missed previous info emails. We could make it just say you can catch up on previous ticket holder updates on the blog and put a general link to the blog? Maybe it's in an if statement so it only appears after the first ticket holder email goes out each year?

russss:
I'd be happy to just point to the blog, I think.

willselby:
Updated!

PR #2048, opened 2026-06-01
Add additional fields to CFP export

### LINKED ISSUE #2017: CSV exports should give a provide columns for link and ID
It would be amazing if the extremely useful CSV export function could provide a link or even CFP ID in the table export.

This would allow for better cross-referencing with other team-members and current grist sources, 
This way we can click through to the CFP in question from the CSV.

jellybob:
- Proposal ID
- Whether the proposing user has a ticket

Closes #2017


marksteward (approved):

chaos-dotcom:
thanks so much!

PR #2062, opened 2026-06-05
Hire stuff has type "rental"

SamLR:
I think these are the places where we care about `ProductGroup.type` which doesn't have a hire _type_ (the `type` is rental, the name is `hire` though...)


SamLR:
in favour of using hire everywhere https://github.com/emfcamp/Website/pull/2063

PR #2063, opened 2026-06-05
Use 'hire' rather than 'rental'

SamLR:
This is the only place I can see that we use 'rental' (and I assume this dict ultimately drives the choices in the productGroup type dropdown)


marksteward (approved):

SamLR:
The only place that I _think_ `PRODUCT_GROUP_TYPES_DICT` is used is [here](https://github.com/emfcamp/Website/blob/main/models/basket.py#L239) where I think it always resolves to `Purchase` for rentals anyway

SamLR:
That being said I can't figure out where it populates the dropdown

marksteward:
It uses PRODUCT_GROUP_TYPES for that (which PRODUCT_GROUP_TYPES_DICT is derived from)

PR #2068, opened 2026-06-07
Volunteer self checkin

jellybob:
UI to allow volunteers to check themselves in for a shift. Once checked in they'll be shown the role notes to provide some guidance on what it is they're meant to be doing.

Role admins now have the ability to edit those notes so they can be evolved during the event as new things are discovered.

Designed to only be accessible via QR codes posted in volunteer venues so it's not linked from anywhere.

### DIFF
--- a/apps/volunteer/role_admin.py
+++ b/apps/volunteer/role_admin.py
@@ -36,7 +36,7 @@ from models.volunteer.shift import (
 class RoleForm(Form):
     name = StringField("Role Name", [InputRequired()])
     full_description_md = StringField(
-        "Description (supprts markdownn)",
+        "Description (supports markdown)",
         [InputRequired()],
         widget=TextArea(),
         description="Displayed in the main role list when signing up.",
--- a/apps/volunteer/role_admin.py
+++ b/apps/volunteer/role_admin.py
@@ -16,7 +16,11 @@ from flask import (
 from flask.typing import ResponseReturnValue
 from flask_login import current_user
 from sqlalchemy import select
+from wtforms import StringField, SubmitField
+from wtforms.validators import URL, InputRequired, Optional
+from wtforms.widgets import TextArea
 
+from apps.common.forms import Form
 from apps.volunteer import v_user_required, volunteer
 from main import db, get_or_404
 from models.user import User
@@ -27,6 +31,28 @@ from models.volunteer.shift import (
     ShiftEntryState,
     ShiftEntryStateException,
 )
+
+
+class RoleForm(Form):
+    name = StringField("Role Name", [InputRequired()])
+    full_description_md = StringField(
+        "Description (supprts markdownn)",
+        [InputRequired()],
+        widget=TextArea(),
+        description="Displayed in the main role list when signing up.",
+    )
+    role_notes = StringField(
+        "Role notes (supports markdown)",
+        [Optional(), InputRequired()],
+        widget=TextArea(),
+        description="Displayed to volunteers when they check in for a shift.",
+    )
+    instructions_url = StringField(
+        "Instructions URL",
+        [Optional(), URL()],
+        description="A link to external instructions. Won't be displayed if role_notes are provided.",
+    )
+    save = SubmitField("Update")
 
 
 @volunteer.route("/role-admin", methods=["GET"])
@@ -116,6 +142,23 @@ def role_admin(role_id):
         offset=offset,
         limit=limit,
     )
+
+
+@volunteer.route("role/<int:role_id>/edit", methods=["GET", "POST"])
+@role_admin_required
+def role_edit(role_id: int) -> ResponseReturnValue:
+    """Allows editing details of a role."""
+    role = get_or_404(db, Role, role_id)
+    form = RoleForm(request.form, obj=role)
+
+    if request.method == "POST" and form.validate():
+        form.populate_obj(role)
+        db.session.add(role)
+        db.session.commit()
+        flash("Role details have been updated.")
+        return redirect(url_for(".role_admin", role_id=role_id))
+
+    return render_template("volunteer/role_edit.html", role=role, form=form)
 
 
 @volunteer.route("role/<int:role_id>/set-state/<int:shift_id>/<int:user_id>", methods=["POST"])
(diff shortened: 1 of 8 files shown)


tests/volunteer/test_shfits.py
lukegb, line 1: shfit

jellybob: Also its an empty file because I started thinking about writing some tests and then remembered the precarious stack of fixtures required so I'll just delete it.

apps/volunteer/role_admin.py
```diff
@@ -29,6 +33,28 @@
 )
 
 
+class RoleForm(Form):
+    name = StringField("Role Name", [InputRequired()])
+    full_description_md = StringField(
+        "Description (supprts markdownn)",
```
lukegb: nit: supports

apps/volunteer/schedule.py
```diff
@@ -243,6 +244,155 @@ def shift_cancel(shift_id):
...
+def self_checkin(venue_id: int) -> ResponseReturnValue:
+    """Allow a user to check themselves into a shift at a specific venue.
+
+    Intended to be accessed via QR codes in those venues. When loaded we'll check
+    whether the user has a scheduled shift in this venue either starting within
+    the next 15 minutes or currently in progress and if so allow them to mark
+    themselves as arrived.
+
+    If they're already marked as arrived then they'll be able to mark themselves
+    as either abandoning the shift, or if we're within 15 minutes or shift end
```
lukegb: nit: of

apps/volunteer/schedule.py
```diff
@@ -243,6 +244,155 @@ def shift_cancel(shift_id):
...
+@feature_flag("VOLUNTEERS_SCHEDULE")
+@v_user_required
+def self_checkin_arrived() -> ResponseReturnValue:
+    shift_entry_id = int(request.form["shift_id"])
+    shift_entry = _get_shift_entry_for_user(shift_entry_id, current_user)
+    if not shift_entry.eligible_for_checkin_at(_now()):
+        return redirect_next_or_schedule("Unexpected volunteer in bagging area.")
+
+    shift_entry.set_state(ShiftEntryState.ARRIVED)
+    db.session.add(shift_entry)
```
lukegb: iirc you don't need to `db.session.add` things which were loaded from this session, so this is redundant (here and below)

templates/volunteer/role_edit.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block title %}
+    Edit {{ role.name }}
+{% endblock %}
+
+{% block body %}
+{% include "volunteer/_nav.html" %}
+
+    <h2>{{ role.name }}</h2>
+    <div class="well">
```
lukegb, line 12: Since these are editable we should make sure we export them at the end so we can update the fixtures for next time.

templates/volunteer/self_checkin.html
```diff
@@ -0,0 +1,101 @@
...
+    <h3>Open shifts</h3>
+    <p>Click the relevant button to check in or out of these active shifts.</p>
+    <table class="table table-bordered responsive-table shifts-table checkin">
+        <thead><tr>
+            <th>Starts</th>
+            <th>Ends</th>
+            <th>Role</th>
+            <th></th>
+        </tr></thead>
+        <tbody id="tbody-{{ active_day }}">
```
lukegb: Does `active_day` get set anywhere? I think this is just a holdover from the schedule logic?

templates/volunteer/self_checkin.html
```diff
@@ -0,0 +1,101 @@
...
+            {% for shift_entry in upcoming_shift_entries %}
+                {% set shift = shift_entry.shift %}
+                <tr id="shift-{{ shift_entry.id }}">
+                    <td class="start_time">{{ shift.start.strftime("%H:%M") }}</td>
+                    <td class="end_time">{{ shift.end.strftime("%H:%M") }}</td>
+                    <td class="time mobile-only">
+                        {{ shift.start.strftime("%H:%M") }}{% if shift.end %} to {{ shift.end.strftime("%H:%M") }}{% endif %}
+                    </td>
+                    <td class="role">{{ shift.role.name }}</td>
+                    </td>
```
lukegb: Duplicate `</td>`

templates/volunteer/self_checkin.html
```diff
@@ -0,0 +1,101 @@
...
+                    <td class="button">
+                        {% if shift_entry.eligible_for_checkin_at(now) %}
+                            <form method="post" action="{{ url_for('.self_checkin_arrived', next=url_for('.self_checkin', venue_id=venue.id)) }}">
+                                <input type="hidden" name="shift_id" value="{{ shift.id }}">
+                                <button class="btn btn-block btn-success">Check in</button>
+                            </form>
+                        {% elif shift_entry.eligible_for_checkout_at(now) %}
+                            <form method="post" action="{{ url_for('.self_checkin_abandon', next=url_for('.self_checkin', venue_id=venue.id)) }}">
+                                <input type="hidden" name="shift_id" value="{{ shift.id }}">
+                                <button class="btn btn-block btn-danger">Mark Abandoned</button>
```
lukegb, line 57: We should possibly contemplate some JS to ask if they're positive they didn't complete this shift, but I dunno how much this is going to be a problem. Mostly just worried about the two buttons being quite close to each other, and most people will be doing this on mobile.

apps/volunteer/schedule.py
jellybob: Oh, this is good information to know. Very much cargo culted that one.

templates/volunteer/self_checkin.html
marksteward: Better to make it easy to fix if/when someone makes a mistake.

jellybob: Agreed, we now give people the ability to undo marking themselves as abandoned or checked out, and as a side effect of that table existing now they can also see if they've been marked as a no show and see previous shifts in this location.

templates/volunteer/role_edit.html
jellybob: #2083

models/volunteer/role.py
marksteward, line 30: This should be versioned if people are going to edit it during the event. I think just:

```
__versioned__: dict[str, Any] = {}
```

marksteward (changes requested):

PR #2073, opened 2026-06-10
Use .table class on wiki tables

ConnorMcF:
This PR makes tables on the wiki use the `table.table` class so they look a little less cramped.

image
image


marksteward:
Thanks! Removed from my list for this evening!

Issue #2075, opened 2026-06-12
Allow villages to disable join requests


russss:
Many villages don't want random people joining their village. There needs to be an option to disable the join request button.

marksteward:
Fixed in #2077

PR #2077, opened 2026-06-12
adds private villages toggle, preventing join requests

ab-gh:
Adds a 'private village' toggle, disabling join requests for a village. #2075

- bool field `private` added to Villages model
- added condition `not village.private` to user_can_join_village
- adds checkbox and explanation to forms
- migration

If a village is set to Private, an authenticated village-less user will not see the join button:
image

And will see an error banner if they visit /members/join directly 
image

Village Admins see a checkbox for Private flag
image


templates/villages/register.html
```diff
@@ -1,4 +1,4 @@
-{% from "_formhelpers.html" import render_field %}
+{% from "_formhelpers.html" import render_field, render_checkbox %}
```
lukegb: Oh. We should probably import _formhelpers.html into the form, rather than into the templates that include them (oops).

ab-gh: i didn't wanna change too much but yeah, i can include that in this PR if that's useful

ab-gh: resolved [b5b5aa6](https://github.com/emfcamp/Website/pull/2077/commits/b5b5aa6126df4cf1183a37430481df5fbab34603)

lukegb (approved):

PR #2079, opened 2026-06-13
volunteers: add public dashboard

Kunsi:
Screenshot 2026-06-13 at 14 21 32

### DIFF
--- a/apps/volunteer/schedule.py
+++ b/apps/volunteer/schedule.py
@@ -79,6 +79,32 @@ def redirect_next_or_schedule(message: str | None = None) -> ResponseReturnValue
     return redirect(next)
 
 
+@volunteer.route("/public-dashboard")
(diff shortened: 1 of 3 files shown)


jellybob:
Copied from a conversation on Mattermost:

If this is intended for big screens, which I think sounds like a good idea, I'd probably do the following:

- Put the emphasis on volunteer role rather than location. It's possible we don't need to show the location at all.
- Larger text for the time & people needed.
- Display people needed as something like 2/8 people needed.
- Put a big "Register at emfcamp.org/volunteer" at the top so that people can see how to sign up for shifts rather than showing up at the volunteer desk.
- A possibly useful stretch goal here would be to default to only showing shifts which are below their minimum volunteer requirements, and only if none of those exist (or there's less than a threshold of them) show shifts that are below their maximum requirements.

marksteward (approved):

PR #2091, opened 2026-06-15
Put a list on /about/contact somewhere with all the EMF fedi accounts

lukegb:
Draft because I need to confirm that someone is actually looking at all of these if we're going to publish them.


marksteward:
Is it worth removing the @emfcamp.org if they've all got it?

lukegb:
Hmm, maybe, although it's slightly harder to copy and paste in a format that's usable in a fedi client at that point.

marksteward (approved):

PR #2092, opened 2026-06-15
Volunteer shift templates

jellybob:
Add support for Shift Templates which allow regenerating shifts for a role at will by role admins.

Volunteer/role admins can now see the volunteer schedule at all times to allow tweaking shift timings before we open up volunteering to attendees. Once signup has been opened attendees will only see shifts for finalised roles, allowing us to open up shifts earlier than we could otherwise by not blocking on everybody being sorted.


models/volunteer/shift.py
```diff
@@ -112,15 +112,176 @@ def eligible_for_completion_at(self, now: datetime) -> bool:
...
+    @property
+    def end_date(self) -> date:
+        return self.start_date if self.start_time < self.end_time else self.start_date + timedelta(days=1)
+
+    @property
+    def end_datetime(self) -> datetime:
+        return event_tz.localize(datetime.combine(self.end_date, self.end_time))
+
+    @property
+    def shift_start_times(self) -> list[datetime]:
```
jellybob: @ormiret I think I've correctly understood and applied the logic you came up with here, would you be able to give it a once over and check that's true?

apps/volunteer/schedule.py
```diff
@@ -106,6 +105,25 @@ def public_dashboard():
...
+    # Days outside of the permitted range are clamped to that range.
+    if day < permitted[0]:
+        return permitted[0]
+
+    if day > permitted[1]:
+        return permitted[1]
+
+    # We're at a festival. If its before 4am it's not tomorrow yet, show what
+    # is technically the previous day.
+    if now.hour < 4:
```
marksteward: The content system ([apps/cfp/date.py](https://github.com/emfcamp/Website/blob/6b27c32a9329700ad9cd95b2bfb859569106788b/apps/cfp/date.py)) has:

```
CONTENT_DAY_START = time(5)
```

Might be worth aligning, or having this in a similar date.py?

jellybob: Nice, I hadn't seen that. I've not fully standardised here but we are now running 5am-5am everywhere, created #2169 to fully standardise event dates.

marksteward:
Happy to merge this to get the DB change in if you're done with it. Some observations from a quick play around:
 - shifts probably need a minimum duration (e.g. 30 mins) or we need a sanity check limit on the number of shifts for a role
 - durations are parsed as minutes, not hours
 - I guess you know the signup page shows the now-UTC timestamps as if they were naive local ones

jellybob:
I think we're good - durations are intentionally in minutes as I think most people can think of whole minutes slightly better than they can fractional hours, and unfortunately we will sometimes need sub one hour shifts. Sanity checking is a good point, I've added a shift count for each role in role admin which isn't perfect but should at least flag anything wildly out of whack.

Good catch on the signup page times - I'd misunderstood `event_tz.localize` so they were in fact being displayed in the wrong timezone everywhere, I think I've fixed that but if you could have a skim of the last commit to be pushed that would be a good idea because I properly rushed that one through before work.

PR #2098, opened 2026-06-16
Deduplicate vevent uids

### LINKED ISSUE #2097: Schedule Favourites ICal .ics file duplicates UIDs
When attaching an ics ICal file to google calander (no others tested yet) the calander only displays the last scheduled event time even though multiple time slots may be availble.

Each discrete VEVENT for the same logical schedule item contains the same UID, which is causing google calander to render only the last event in the list

Recommend adding a suffix e.g. `2026-123-1`, `2026-123-2`etc...

for context, pulled from my own favourites ics file (truncated for readability)
```
...

BEGIN:VEVENT
SUMMARY:Karaoke
DTSTART;TZID=Europe/London:20260716T193000 <------------ Different Dates/Times
DTEND;TZID=Europe/London:20260716T233000
...
UID:2026-189 <------------------------------------------ See Here
DESCRIPTION:Karaoke returns to EMF...
END:VEVENT

BEGIN:VEVENT
SUMMARY:Karaoke
DTSTART;TZID=Europe/London:20260717T193000 <------------- Different Dates/Times
DTEND;TZID=Europe/London:20260717T233000
...
UID:2026-189 <------------------------------------------ See Here
DESCRIPTION:Karaoke returns to EMF...
END:VEVENT

BEGIN:VEVENT
SUMMARY:Karaoke
DTSTART;TZID=Europe/London:20260718T193000 <------------ Different Dates/Times
DTEND;TZID=Europe/London:20260718T233000
...
UID:2026-189 <------------------------------------------ See Here
DESCRIPTION:Karaoke returns to EMF...
END:VEVENT

BEGIN:VEVENT
SUMMARY:Karaoke
DTSTART;TZID=Europe/London:20260719T193000 <------------ Different Dates/Times
DTEND;TZID=Europe/London:20260719T233000
...
UID:2026-189 <------------------------------------------ See Here
DESCRIPTION:Karaoke returns to EMF...
END:VEVENT

...
```

queekusme:
Add a sequence id to calendar uid entries for the yearly schedule ical as well as user favourites list ical

Resolves #2097

### DIFF
--- a/apps/schedule/feeds.py
+++ b/apps/schedule/feeds.py
@@ -165,12 +165,13 @@ def schedule_ical(year: int) -> ResponseReturnValue:
 
     for flat_sid in flat_sids:
         cal_event = Event()
-        cal_event.add("uid", "{}-{}".format(year, flat_sid["id"]))
+        occurrence = flat_sid["occurrences"][0]
+        cal_event.add("uid", "{}-{}-{}".format(year, flat_sid["id"], occurrence["occurrence_num"]))
         cal_event.add("summary", flat_sid["title"])
         cal_event.add("description", _format_event_description(flat_sid))
-        cal_event.add("location", flat_sid["occurrences"][0]["venue"])
-        cal_event.add("dtstart", flat_sid["occurrences"][0]["start_date"])
-        cal_event.add("dtend", flat_sid["occurrences"][0]["end_date"])
+        cal_event.add("location", occurrence["venue"])
+        cal_event.add("dtstart", occurrence["start_date"])
+        cal_event.add("dtend", occurrence["end_date"])
         cal.add_component(cal_event)
 
     return Response(cal.to_ical(), mimetype="text/calendar")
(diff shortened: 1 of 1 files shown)


russss:
Thanks for helping out! Unfortunately this isn't the right way to fix this, and it looks like the correct way isn't especially straightforward for some reason.

This year we've refactored the schedule system so that one ScheduleItem can have multiple Occurrences. The correct ID to use here is the Occurrence ID, which will be unique.

I *think* that's available in `flat_sid["occurrences"][0]["id"]` but I don't immediately understand what's going on with all this code.

queekusme:
> Thanks for helping out! Unfortunately this isn't the right way to fix this, and it looks like the correct way isn't especially straightforward for some reason.
> 
> This year we've refactored the schedule system so that one ScheduleItem can have multiple Occurrences. The correct ID to use here is the Occurrence ID, which will be unique.
> 
> I _think_ that's available in `flat_sid["occurrences"][0]["id"]` but I don't immediately understand what's going on with all this code.

Hi, I'm still getting around the website code and so am not familiar with all of the intricacies, I'll review the occurrence id and see where it goes

EDIT: since flat_sid["occurrences"][0] is an 'OccurrenceDict' (apps/schedule/data.py#141) I believe 'occurrence_num' may be the unique identifier here (occurrence_num has a unique constraint with schedule_item_id models/contnet/schedule.py#407) 

I'll make the change and resubmit

marksteward:
Yep, the new change should work. A "flat_sid" is a "Schedule Item dict" that contains exactly one occurrence, so the unique ID is either `["occurrences"][0]["id"]` or the combination of `id` and `occurrence_num` as you've done.

The latter is probably better for consistency over cancellation/replacement.

marksteward:
Can you run `ruff format` and squash the commits? Thanks!

queekusme:
> Can you run `ruff format` and squash the commits? Thanks!

done

marksteward (approved):

PR #2100, opened 2026-06-16
Allow converting tickets to tiers which would result in a refund, and process the refund

### LINKED ISSUE #1102: Allow a ticket to be replaced with a cheaper ticket with a refund for the difference
If an accepted speaker already has a ticket and wants to convert it into a discounted ticket, we currently have to handle this off-the-books by issuing a part refund manually. We should be able to do this through the admin interface because this is susceptible to errors and results in a slight difference in the accounting.

We should be able to select a single ticket and replace it with a cheaper ticket, issuing a refund for the difference in price.

lukegb:
This only works for Stripe-backed payments at the moment.

Fixes #1102

### DIFF
--- a/templates/admin/tickets/ticket-convert.html
+++ b/templates/admin/tickets/ticket-convert.html
@@ -21,13 +21,27 @@
 <dt>Capacity</dt><dd>{{ coalesce(new_tier.parent.capacity_max) }}</dd>
 </dl>
 
+{% if price_change %}
+<div class="alert alert-danger">This ticket type conversion will result in a refund of {{ price_change }} {{ticket.price.currency }}.</div>
+{% endif %}
+
 <form method="post">
 {{ form.hidden_tag() }}
+{% if price_change %}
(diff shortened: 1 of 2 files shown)


templates/admin/tickets/ticket-convert.html
```diff
@@ -52,7 +66,12 @@ <h4>Choose a price tier to convert to</h4>
             <td>{{coalesce(tier.parent.capacity_max)}}</td>
             <td>
                 {% if tier != ticket.price_tier %}
-                <a class="btn btn-default" href="{{ url_for('.convert_ticket', ticket_id=ticket.id, price_tier_id=tier.id)}}">Convert</a>
+		{% set will_refund = tier.get_price(ticket.price.currency).value != ticket.price.value %}
```
marksteward: Tabs around here should be spaces

marksteward (approved):
This looks sensible!

PR #2108, opened 2026-06-18
Show talk and workshop descriptions inline on the schedule page

smagdali:
The existing line-up lists every item's speaker and title but hides the description behind a click into each detail page. Better UX to render the description inline instead, so the whole programme is readable on one page. The total page size is actually still manageable (with all the descriptions added, the total page weight increases from 330k to 584k) , but much easier to explore.

Each item becomes a container div with the title as the link (descriptions contain urlized links, so the item can no longer be wrapped in a single anchor), and the description is shown with the same markup as the detail page. The favourite button stays inside the form.

I left the links to the detail page - you might argue that they're redundant now but this way you can still still share a link to a specific event. 

 Based on the first pass I did at https://whitelabel.org/emf-2026/ before I realised this repo was available. 
 
 Disclaimer - I made this with Claude Code.

### DIFF
--- a/css/_lineup.scss
+++ b/css/_lineup.scss
@@ -9,24 +9,21 @@
     visibility: visible;
   }
   .event {
-    text-decoration: none;
-  }
-  .event > div {
-    padding: 10px 4px;
+    padding: 10px 4px 14px;
     border-bottom: 1px solid rgba(100, 100, 100, 0.4);
-    padding-top: 14px;
   }
-  .event:first-child > div {
+  .event:first-child {
     border-top: 1px solid rgba(100, 100, 100, 0.4);
   }
-  .event:hover,
-  .event:focus,
-  .event:active,
-  .event:hover div,
-  .event:focus div,
-  .event:active div {
-    background-color: #33071c;
-    color: inherit;
+  .schedule-item-description {
+    max-width: 70ch;
+    margin: 0.4em 0 0;
+  }
+  &.descriptions-collapsed .schedule-item-description {
+    display: none;
+  }
+  [data-toggle-descriptions] {
+    margin-bottom: 10px;
   }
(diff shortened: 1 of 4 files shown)


marksteward:
Thanks for this, I think it might be a more useful fix for #2115. However, I don't think we want to make this the default - it would need to be behind a toggle button or separate "full programme" page.

You seem to have removed all visual feedback on the links, and included redundant `text-decoration: underline` for a link class. Were these changes to fix something affecting your browser, or just something Claude mangled?

And please remove the `Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>`, we don't want to be part of a move by Anthropic to prove their tools have authorship rights. I'm not sure how Anthropic's terms mesh with our [licence](https://github.com/emfcamp/Website/blob/main/LICENSE), and it would be better to be clear about whether you're claiming ownership of these changes. If you used Claude to assist you, that's fine, but if it's vibe-coded I think it's safest to treat this as a detailed feature request instead.

smagdali:
1. Descriptions are now opt-in behind a "Show descriptions" toggle on the line-up. They're collapsed by default, the choice is remembered per browser (localStorage), and the per-user shuffle in line_up() is untouched. Without JavaScript the descriptions are shown and the toggle is hidden, so nothing is lost for no-JS users.

2. Link styling. I didn't fully understand your comment at first. I actually thought it was better to bring the link styling to be more consistent with the rest of the site, and follow best UX practice that links should be distinguishable from non-links without a mouseover. But I added the underline without the highlight colour change-  it should work correctly now.

3. Claude Authorship trailer removed.  I own this, not that fancy linter. rebased to remove the original claim.

marksteward (approved):

marksteward:
Looks good, thanks!

PR #2109, opened 2026-06-18
Add info on how to set a village marker on the map

ConnorMcF:
Seen several people asking how to do this and it isn't super clear, so a little text on the villages page might be useful.


russss:
If you've got a sec it would be good to add a link to the "edit village" page as well

PR #2111, opened 2026-06-18
Add Fediverse attribution

ConnorMcF:
Not very important but why not, highlight the EMF fedi accounts a little more :^)

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -17,6 +17,7 @@
         <meta property="og:description" content="{{ self.description()|replace("\n", " ") }}">
         <meta name="description" content="{% filter replace("\n", " ") %}{% block description %}{% endblock %}{% endfilter %}">
         {% endif -%}
+        <meta name="fediverse:creator" content="@emf@emfcamp.org">
         <meta name="theme-color" content="#171123">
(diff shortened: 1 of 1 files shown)


templates/base.html
```diff
@@ -17,6 +17,8 @@
         <meta property="og:description" content="{{ self.description()|replace("\n", " ") }}">
         <meta name="description" content="{% filter replace("\n", " ") %}{% block description %}{% endblock %}{% endfilter %}">
         {% endif -%}
+        <link href="https://social.emfcamp.org/@emf" rel="me">
```
marksteward: I'm not sure rel=me is needed here, I don't want us to accidentally opt into some sort of verified post or RelMeAuth thing and it's all on the same domain already.

marksteward (changes requested):

templates/base.html
ConnorMcF: It just gives you a little green verified tick for emfcamp.org
image
Like this:
image

If there's concerns about other stuff like RelMeAuth (does anything use that?) I can take it out. I think it's not a problem personally.

ConnorMcF: Swapped it to `fediverse:creator` only

marksteward: I think that might only be on Mastodon. Our social account is hosted at https://social.emfcamp.org/@emf which is GoToSocial

PR #2114, opened 2026-06-18
Fix date format in nav

nbookham:
The formatting was bugging me :)

### DIFF
--- a/templates/_nav.html
+++ b/templates/_nav.html
@@ -7,7 +7,7 @@
   <div id="header-text">
     Eastnor, UK
     <br>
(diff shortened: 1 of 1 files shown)


Kunsi:
i think this is duplicated for the public volunteer dashboard, so maybe also fix it there?

lukegb:
Hmm, this is a matter of taste rather than correctness, I fear.

russss:
Yeah this was a style choice and while I might be inclined to revisit it, we don't really have the capacity to discuss this stuff at this stage.

Issue #1499, opened 2024-05-07
Add a next button to browse through lineup


SamLR:

Jonty:
Specifically this is on the proposal view page, so you can go to the next one without having to go back to the list while skimming through

marksteward:
Duplicate of #447

Issue #447, opened 2016-07-31
Add next/previous links to line-up items


davidc:
When viewing a specific talk (/line-up/2016/37-a-call-to-arms-lets-hack-usb-devices), it would be nice to have next/previous links that take you to the next/previous in the same order that the /line-up/2016 page was listed.

For people like me who are reading all the descriptions to decide which to attend, to avoid having to go back to the main list after each.

marksteward:
Fixed by #2115

PR #2115, opened 2026-06-19
Add next/prev button to line-up

marksteward:
Adds the buttons requested in #447, #1499 and on IRC. Only works when we're at the line-up stage, not when the schedule is fully enabled.

Unfortunately, people can come into the item page in 3 ways:
 - line-up
 - favourites
 - direct link

This PR assumes the only way in is via the line-up. Visitors to a shared link will see unexplained Next/Previous buttons, and people clicking on their Favourites a few days later will find that Next/Previous follow the line-up instead of the (undefined) order of their favourites.

I've considered setting a cookie when hitting the line-up page and clearing it on favourites but I think that will be just as confusing. Any suggestions?

### DIFF
--- a/apps/schedule/base.py
+++ b/apps/schedule/base.py
@@ -2,6 +2,7 @@ import random
 from collections import defaultdict
 from datetime import datetime, timedelta
 from itertools import groupby
+from typing import Any
 
 import pendulum
 from flask import abort, flash, redirect, render_template, request, url_for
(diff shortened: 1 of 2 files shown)


russss:
I'm not especially keen on using a cookie. I think a URL parameter is a better solution here. We could put the user ID in the URL parameter so it only shows next/previous for that user.

lukegb:
I kinda think we're maybe overthinking it, and showing next/prev is fine?

Ideally we don't want people to think we've added YouTube-style `si` share tracking parameters even if we aren't using it for that...

russss:
Yeah I didn't spot this was only shown when SCHEDULE wasn't enabled. I think this is fine.

russss (approved):

PR #2117, opened 2026-06-19
Add event opening and closing to favourites.ics

edk0:
Also don't include the actual schedule until SCHEDULE is enabled

### DIFF
--- a/apps/schedule/feeds.py
+++ b/apps/schedule/feeds.py
@@ -1,5 +1,7 @@
 import json
+from datetime import timedelta
 
+from dateutil.parser import parse as parse_date
 from flask import Response, abort, redirect, request, url_for
 from flask import current_app as app
 from flask.typing import ResponseReturnValue
@@ -14,7 +16,7 @@ from models.user import User
 
 from ..common import feature_enabled, feature_flag, json_response
 from ..config import config
-from . import schedule
+from . import event_tz, schedule
 from .data import (
     ScheduleFilter,
     ScheduleItemDict,
@@ -166,7 +168,7 @@ def schedule_ical(year: int) -> ResponseReturnValue:
     for flat_sid in flat_sids:
         cal_event = Event()
         occurrence = flat_sid["occurrences"][0]
-        cal_event.add("uid", "{}-{}-{}".format(year, flat_sid["id"], occurrence["occurrence_num"]))
+        cal_event.add("uid", f"{year}-content-{flat_sid['id']}-{occurrence['occurrence_num']}")
         cal_event.add("summary", flat_sid["title"])
         cal_event.add("description", _format_event_description(flat_sid))
         cal_event.add("location", occurrence["venue"])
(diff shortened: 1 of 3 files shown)


apps/schedule/feeds.py
```diff
@@ -220,23 +222,43 @@ def favourites_ical() -> ResponseReturnValue:
...
     cal.add("version", "2.0")
 
+    fixed_events = app.config.get("FIXED_EVENTS", {})
+    event_address = app.config.get("EVENT_ADDRESS", f"EMF {config.event_year}")
+
+    for i, fev in enumerate(fixed_events.items()):
+        name, start_s = fev
+        start_date = event_tz.localize(parse_date(start_s))
+        cal_event = Event()
+        cal_event.add("uid", f"{config.event_year}-x{i}")
```
marksteward: I don't love `x` here, maybe we could add an extra prefix `fixed` vs `content`, and update https://github.com/emfcamp/Website/blob/bce08ee1941ce900191bd85cbf4b19a85b59ef51/apps/schedule/feeds.py#L169

config/orchestrator.cfg
```diff
@@ -108,6 +108,13 @@ RESERVE_LIST_TICKET_LINK = "http://www.emfcamp.org/tickets/token/invalid"
 EVENT_START = '2026-07-16 11:00:00'
 EVENT_END   = '2026-07-19 23:00:00'
 
+FIXED_EVENTS = {
+  "Gates open": EVENT_START,
+  "Site closes": '2026-07-20 12:00:00',
```
marksteward: It'd be nice to harmonise this with https://www.emfcamp.org/about/arrival-times in the future, so maybe we call these something like GATE_OPEN and GATE_CLOSED, instead of having the labels in a config file.

apps/schedule/feeds.py
```diff
@@ -220,23 +222,43 @@ def favourites_ical() -> ResponseReturnValue:
...
+        cal_event.add("summary", name)
+        cal_event.add("location", event_address)
+        cal_event.add("dtstart", start_date)
+        cal_event.add("dtend", start_date + timedelta(minutes=1))
```
marksteward: I usually find calendar apps do silly things with 1-minute event lengths. I think an hour might be more suitable, because it's not going to be accurate to the minute anyway.

apps/schedule/feeds.py
```diff
@@ -220,23 +222,43 @@ def favourites_ical() -> ResponseReturnValue:
...
+    title = f"EMF {config.event_year} Favourites for {user.name}"
 
     cal = Calendar()
     cal.add("summary", title)
     cal.add("X-WR-CALNAME", title)
     cal.add("X-WR-CALDESC", title)
     cal.add("version", "2.0")
 
+    fixed_events = app.config.get("FIXED_EVENTS", {})
+    event_address = app.config.get("EVENT_ADDRESS", f"EMF {config.event_year}")
```
marksteward: This probably shouldn't default to the name of the event. You can do `app.config['EVENT_ADDRESS']` instead of `.get`.

marksteward (changes requested):

Issue #1926, opened 2026-03-23
Add a ProposalType object?


marksteward:
#1162 and #1107 would benefit from somewhere to store state like "we're still accepting workshops", and "talks should be 45 minutes long". This could be somewhere to migrate some of PROPOSAL_INFOS to, and the awful `feature_enabled(f"CFP_{proposal_type.upper()}S_CLOSED")`.

It might also nice way to formalise the logic that lottery wins cancel all other entries within a type.

To be clear, I don't want us to be able to add new ProposalTypes on the fly because of there's so much template and view logic tied to them. We should keep the attribute information in code for now.

marksteward:
We should move LINEUP_TYPE_ORDER into here, and make it clear it's also used as a per-type publishing flag.

Also:
 - #665
 - #1483 (at least cfp admin perms in general)

PR #2122, opened 2026-06-21
Limit line-up APIs to LINEUP_TYPE_ORDER

edk0:


marksteward:
Thanks, merging on the assumption that #1926 will fix the now misleading naming.

PR #2123, opened 2026-06-22
Update attendee content wording

mscroggs:
submitted -> accepted

### DIFF
--- a/templates/schedule/attendee_content/index.html
+++ b/templates/schedule/attendee_content/index.html
@@ -7,7 +7,7 @@
     adding them to the schedule.
 </p>
 <p>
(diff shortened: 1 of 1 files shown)


marksteward (approved):

russss:
I think if someone's content has been accepted to the main CfP, then they're probably not going to submit it as village content. (And if they want to do it again at a village, we'd welcome that! But this wording would probably discourage them from doing it.)

I think the intent of this message was to prevent people confusing the village content with the main CfP, so it's probably worth rewording this in a different way.

russss:
I ended up making a slightly different change in da67db7d

PR #2124, opened 2026-06-23
Add Music and Meetup Schedule item types

### LINKED ISSUE #1515: Add music type to proposals
This only needs the minimum information to get it scheduled

* Title
* Description
* Scheduled time

Jonty:
This allows both us and attendees to create and schedule Music and Meetup types.

This was _way_ easier to implement than I expected - a+ content system refactoring. It all works fine, but let me know if I missed something here?

Fixes #1515

### DIFF
--- a/apps/cfp_review/forms.py
+++ b/apps/cfp_review/forms.py
@@ -247,6 +247,14 @@ class UpdateScheduleItemFilmForm(UpdateAttributesForm):
     classification = StringField("Classification")
 
 
+class UpdateScheduleItemMusicAttributesForm(UpdateAttributesForm):
+    pass
+
+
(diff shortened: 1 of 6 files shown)


russss:
This works standalone and fixes most of our issues so I'll merge it, but it is missing one part: if a ScheduleItem is derived from a Proposal, then we won't be able to convert it to the "music" type.

This is because the admin interface for converting a CfP-derived ScheduleItem actually sends you to convert the corresponding Proposal, with the conversion then cascading down to the ScheduleItem. (Hence my three options in #1515.)

I'll create a new issue for that.

Jonty:
Ugh, I didn't hit that in testing - must have been looking at a non-proposal-linked scheduleitem. I'll see if I can fix that too.

PR #2134, opened 2026-06-24
Add shortlink to phonebook

ConnorMcF:
Makes the QR code a little less dense!

### DIFF
--- a/apps/base/__init__.py
+++ b/apps/base/__init__.py
@@ -143,6 +143,11 @@ def phones() -> ResponseReturnValue:
     return redirect("/about/phones")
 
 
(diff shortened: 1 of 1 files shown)


edwinbalani:
We could use the emf.camp shortlink system _if_ that counts as perminfra.  (Most, but not all, of the stuff under emf.camp isn't perminfra)

ConnorMcF:
As in https://emf.camp/phonebook? If so that would be ideal

lukegb:
That's perminfra. Just edit the grist: https://grist.orga.emfcamp.org/mFNdD6hhfmaS/URL-Shortener

PR #2137, opened 2026-06-24
Correct estimation by removing unnessecary changover

Jonty:
We place timeblocks to ensure that any changeover between content types is taken into account between them. By counting changeover time on every occurrence we over-estimate the time needed to place them.

This subtracts $changeover_period * blocks from the amount of time allocated in order to correct for the over-estimate.


Jonty:
This really needs a VERY GOOD eye over and some thinking before merge.

PR #2154, opened 2026-06-28
Make motorbike parking not needing a ticket clearer

ConnorMcF:
Seen several people ask this, so making it clearer on the travel page.

### DIFF
--- a/templates/about/travel.md
+++ b/templates/about/travel.md
@@ -26,7 +26,7 @@ Timetables and further information can be found on the [shuttle bus page](https:
 #}
 
 ## By Car
-If you plan to drive, you will need to [buy a parking ticket or live-in vehicle ticket](/tickets/other). Motorcycles do not require a parking ticket. 
+If you plan to drive, you will need to [buy a parking ticket or live-in vehicle ticket](/tickets/other).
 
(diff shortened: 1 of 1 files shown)


templates/about/travel.md
```diff
@@ -36,6 +36,10 @@ Unless you are travelling in a live-in vehicle, you will be parked in the main c
 
 We encourage attendees to carpool where possible. There's a [lift sharing](https://www.emfcamp.org/wiki/lift-sharing) page to find other people on the same route.
 
+### Motorcycles
```
marksteward: Sorry to be annoying but I think this would be better as a level-2 heading after the EV section, like "By Bike" below.

templates/about/travel.md
```diff
@@ -36,6 +36,10 @@ Unless you are travelling in a live-in vehicle, you will be parked in the main c
...
+
+Motorcycles do not require a parking ticket.
```
marksteward: Suggest putting "motorbikes" for the second instance of motorcycles to make it easier to search the page

marksteward (changes requested):

marksteward (approved):

marksteward:
Thanks!

PR #2170, opened 2026-07-02
Only installations still open

katiesteckles:


marksteward:
I've replaced this with 3b9684de which works it out automatically.

PR #2185, opened 2026-07-04
Add support for Google Wallet pass creation

lukegb:
This mimics the PKPass generation in some ways, but I didn't replicate certain aspects (e.g. the automatic generation of images from brand imagery).

For Google Wallet we support updating passes on-the-fly with an admin command - in principle the same thing is possible for Apple Wallet but that only works for users who are using the 'actual' Wallet implementation (users of third-party apps wouldn't benefit). We need to be able to generate the pass class using an admin command, which needs to be done whenever the definition changes - we could probably automate this at startup of the container but for now I haven't bothered, since it shouldn't change _too_ often...


marksteward:
Looks good (this was downloading in a browser):

Screenshot_20260704-174030
Screenshot_20260704-174047

You can't generate a class without a publicly accessible URL.

marksteward (approved):

PR #2191, opened 2026-07-05
Add table and UI for schedule item presenters

marksteward:
Adds presenter management UI.


marksteward:
Rebase and merged as 6aaf34f8

PR #2198, opened 2026-07-05
Add donors

joenash:
Adding two of the recent donors, and a special in-kind sponsor via Jonathan


templates/sponsors/sponsors.html
```diff
@@ -85,12 +85,23 @@ <h3 class="sponsor-tier-name sponsor-tier-name--badge">Badge</h3>
   </div>
 </div>
 
+<div class="row sponsor-row">
+  <div class="clearfix col-xs-2"></div>
```
lukegb: This got mangled a bit, I think vs. the other ones? It should be col-xs-12 and should wrap the other divs.

lukegb (changes requested):

templates/sponsors/sponsors.html
joenash: I have no earthly idea what happened here, the consequence of doing this in a meeting. Shall adjust!

templates/sponsors/sponsors.html
```diff
@@ -85,12 +85,24 @@ <h3 class="sponsor-tier-name sponsor-tier-name--badge">Badge</h3>
...
+</div>
+
 <div class="row sponsor-row">
   <div class="clearfix col-xs-12">
     <h3 class="sponsor-tier-name sponsor-tier-name--copper">Copper</h3>
     <ul>
       <li>Pimoroni</li>
       <li>Hookdeck</li>
+      <li>Dee "iVPN is better anyway"</li>
+      <li>🏳️‍⚧️Creatures for Civility, Cameradery, Community, and human
```
marksteward, line 104: Is this meant to be Camaraderie?

joenash: That is the text as the donor sent it! Probably safe to assume though

PR #2200, opened 2026-07-05
Allow shift notes

jellybob:
These get displayed on the volunteer schedule under the role name for cases where we need specific details to be shown, its primarily intended for workshop helpers, where the volunteer requirements are incredibly variable from shift to shift.


jellybob:
This one has an attached database migration so I'll leave it to be merged when someone who can do that is around.

marksteward (approved):

PR #2206, opened 2026-07-06
Receipt changes

marksteward:
There's enough in here that it could use an eye over

### DIFF
--- a/apps/common/walletpass.py
+++ b/apps/common/walletpass.py
@@ -4,7 +4,8 @@ import hashlib
 import io
 import json
 import zipfile
-from collections.abc import Iterable
+from collections import defaultdict
+from collections.abc import Iterable, Mapping
 from datetime import datetime
 from functools import lru_cache
 from pathlib import Path
@@ -28,7 +29,7 @@ from PIL.Image import Palette, Resampling
 from apps.common.receipt import get_purchase_metadata
 from apps.config import config
 from main import static_digest
-from models.user import User
+from models.user import User, verify_checkin_code
 
 if TYPE_CHECKING:
     pass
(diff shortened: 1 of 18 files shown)


apps/tickets/tasks.py
```diff
@@ -512,99 +523,42 @@ def ticket_url(email):
...
-        app.logger.debug("Google Wallet ticket differences for %s: %s", user.email, differences)
-        app.logger.info("Updating Google Wallet ticket %s for user %s", new_pass["id"], user.email)
-        client.update(resourceId=new_pass["id"], body=new_pass).execute()
+            click.echo(f"User {user.email} has no ticket in Google Wallet's backend.")
+
+        walletpass.update_gwallet_pass(client, user, old_pass, new_pass)
+
+    elif all_users:
+        old_passes_by_user_id = walletpass.get_all_gwallet_passes(client)
+        user_query = db.select(User).options(db.joinedload(User.buildup_volunteer))
```
lukegb, line 554: I don't remember why this joinedload's buildup_volunteer, and it probably shouldn't.

marksteward: It's to correct the arrival time shown on the ticket, I reckon it's good.

apps/common/walletpass.py
```diff
@@ -686,3 +688,113 @@ def gwallet_api_client() -> googleapiclient._apis.walletobjects.v1.resources.Wal
...
+def get_all_gwallet_passes(
+    client: googleapiclient._apis.walletobjects.v1.resources.WalletobjectsResource,
+) -> Mapping[int, list[googleapiclient._apis.walletobjects.v1.EventTicketObject]]:
+    has_next = True
+    next_token = None
+    # I'm not sure if it's possible for multiple tickets to exist for a specific user,
+    # but let's assume it is for some reason, and that we want to update them all.
+    passes_by_user_id = defaultdict(list)
+    while has_next:
+        # God this interface sucks
```
lukegb, line 702: pagination is hard etc

apps/tickets/tasks.py
lukegb: Ah, right, yeah, that makes sense. SGTM.

lukegb (approved):

apps/common/walletpass.py
marksteward: 🫠

PR #2211, opened 2026-07-07
Show the Not Recording icon for review talks too

barnoid:
Discussed briefly on MM content-greenroom. Maybe to be obsoleted by the schedule viewer.

### DIFF
--- a/templates/schedule/_schedule_item_icons.html
+++ b/templates/schedule/_schedule_item_icons.html
@@ -1,6 +1,6 @@
 {# reusable icons #}
 {% macro schedule_item_icons(schedule_item) %}
-  {% if schedule_item.video_privacy == 'none' %}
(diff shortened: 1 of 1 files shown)


templates/schedule/_schedule_item_icons.html
```diff
@@ -1,6 +1,6 @@
 {# reusable icons #}
 {% macro schedule_item_icons(schedule_item) %}
-  {% if schedule_item.video_privacy == 'none' %}
+  {% if schedule_item.video_privacy in ('none', 'review') %}
```
Kunsi: in the json feeds we use `!= "public"`, maybe use that here for consistency?

barnoid: Done

PR #2222, opened 2026-07-08
More volunteer tweaks

jellybob:
Saving this to merge tomorrow once I've reviewed on some sleep.

- Allow setting training notes even if you don't require people to read them so the kitchen can publish the stuff they tell people in person for review.
- Volunteer admins see their own roles on the training page, even if its not one of their chosen volunteer roles.
- People who have a checked in ticket can sign up for shifts on Day 0 even if they're not registered for build, so that early village arrivals/installations/etc can sign up for shifts once they're set up.

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -515,6 +515,11 @@ class User(BaseModel, UserMixin):
         """Whether the user has a ticket to the event."""
         return self.admission_tickets_held > 0
 
+    @property
+    def is_checked_in(self) -> bool:
+        """Whether the user has checked in an admission ticket at the gate."""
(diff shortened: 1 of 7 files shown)


models/user.py
```diff
@@ -515,6 +515,11 @@ def has_admission_ticket(self) -> bool:
         """Whether the user has a ticket to the event."""
         return self.admission_tickets_held > 0
 
+    @property
+    def is_checked_in(self) -> bool:
+        """Whether the user has checked in an admission ticket at the gate."""
+        return any(t.redeemed for t in self.get_owned_tickets(paid=True, type="admission_ticket"))
```
jellybob, line 521: I think this is correct for assessing whether a user has arrived but I'm not 100%.

PR #2241, opened 2026-07-09
Add a button to exit the workshop lottery

boreq:
I think this is useful so that people don't take spots if they change their mind.


jellybob (approved):

boreq:
Seems to work locally.

boreq:
I was thinking about adding a guard to the button to avoid miss-clicks but I think if you exit the lottery and reenter nothing really changes so I think screw it.

jellybob:
Yeah, nothing gets allocated until the lottery closes so its not like people are going to lose tickets by misclicking and then not being able to reenter.

jellybob:
Added to the merge queue, thanks @boreq!

PR #2242, opened 2026-07-09
Remove 'only family friendly' filter

lukegb:


marksteward:
This is actually wanted, just not everyone knew about it (and it's apparently buggy).

PR #2243, opened 2026-07-09
Workshop lottery form only makes sense if user is logged in

lukegb:


jellybob:
This could probably use an alternate branch which suggests people log in if they want to enter the lottery.

jellybob (approved):

jellybob (changes requested):
Regrettably you now have a conflict due to the rate at which things are being merged.

jellybob (approved):

PR #2244, opened 2026-07-09
Fix frab exporter trying to sort rooms with differing methods

lukegb:


Kunsi:
ooops, how did this not explode in dev

PR #2246, opened 2026-07-09
Fix 500 for logged out users

boreq:
If you got to a workshop page logged out (via a link) then you get a 500.


jellybob:
#2243 got there before you :)

PR #2255, opened 2026-07-09
Adjust wording for drop-in workshops

willselby:
Reducing this to a simple "yes" and "no" makes it clearer. Currently it's a bit wordy:

Screenshot 2026-07-09 at 20 38 17


marksteward:
This is a new term this year that's not explained anywhere, I figured I should either add a little callout or icon with an explanation, or just use term in context.

PR #2263, opened 2026-07-10
Update travel.md

trystcat:
Finalised information about the shuttle bus service.


templates/about/travel.md
```diff
@@ -35,6 +37,16 @@ The shuttle bus service will operate between the following times. A full schedul
   </tbody>
 </table>
 
+#### Important ####
```
jellybob: Can you move this up the page above the schedule and make sure it also mentions busses being busy Thursday/Friday and that people should be prepared for a wait?

jellybob (approved):

PR #2269, opened 2026-07-10
only show transferred email to sender

SamLR:
they have to know this but we should minimise leaked information


marksteward:
Rebased and merged as 94d76f99d1c5f9612505b390f4322108c4ed9921

PR #2298, opened 2026-07-12
Add a standard paragraph saying what a drop-in is

mscroggs:


apps/schedule/data.py
```diff
@@ -24,6 +24,12 @@
 from ..config import config
 from . import event_tz
 
+_drop_in_info = (
+    "\n\nThis is a drop-in workshop, so you can arrive at any point during "
```
mscroggs, line 28: Based on other pages, it looks like \n\n should do a paragraph break, but I don't have a website build running locally to check

mscroggs:
Just seen #2255. The suggestion there of adding a drop-in icon, with something like the text here as hover over text might be a better solution

PR #2302, opened 2026-07-12
Fix typo

AlexJLP:


templates/volunteer/training/bar/things-to-know.md
```diff
@@ -12,7 +12,7 @@ People come to the bar asking for this with various regional pronunciations.
 This is another word for Thimble, which refers to the measures used to pour wines and spirits on the back bar.
 
 ## Tschunk
-Pronounced [sh-un-k], tschunk is a cocktail popular with our German hacker crowd, particularly on the last night of EMF.
+Pronounced [ˈtʃʊŋk], tschunk is a cocktail popular with our German hacker crowd, particularly on the last night of EMF.
```
jellybob, line 15: Can you revert this change please (the other one is fine) - while technically correct even at EMF significantly fewer people can read IPA than can sound out some syllables that have been broken up.

marksteward: The spelled out version is also wrong - it should be something like "choonk". Maybe it's better to just cut it.

marksteward: I've fixed this with a massive cop-out 6555923589225c9e0870db3b87be6bc27b6c179a

marksteward:
Cherry-picked and merged as 7f224d1953ddee34597a547c1c0cfdb7957c840e

PR #2309, opened 2026-07-13
Add signal offtopic link

craftbyte:


russss:
I surveyed 4 people in HQ about this change and they all said "no", sorry.

PR #2313, opened 2026-07-14
Edit text

katiesteckles:
We will not be re-opening anything


willselby:
Maybe mention something about being able to queue from 15 minutes before the workshop starts? Or link to https://www.emfcamp.org/schedule/lottery/about for more info?

katiesteckles:
Yeah good shout - I'm also not sure the definition of 'cancelled' was super clear so have simplified.

PR #2324, opened 2026-07-14
Lightning talks fix

SamLR:
This is horrible

It adds a new table `lightning_talk` which keys off an occurrence for a scheduled item called "Lightning Talk". It assumes talks are 5 minutes long.

Each talk has a title, a description and a link for slides. I'm not sure the latter is needed but it's probably better than having them be bought via USB

I've not implemented changing session as dropdowns are the devil (I couldn't make it work). If need-be someone can delete their talk and re-submit (gone with deletion Vs cancel as it keeps things simple)

The list of lightning talks can be accessed via `cfp/proposals` (which is one of the account menus).

I think this works but please test it to confirm. 

There is currently no back end view


marksteward (approved):
This looks great!

templates/cfp/edit_lightning_talk.html
```diff
@@ -0,0 +1,31 @@
+{% from "_formhelpers.html" import render_field %}
+{% from "cfp/_proposal_type_menu.html" import proposal_type_menu with context %}
+{% extends "base.html" %}
+{% block title %}Call for Participation{% endblock %}
+{% block description %}Electromagnetic Field is looking for lightning talks for
+our non-profit festival in the UK this summer.{% endblock %}
+{% block body %}
+    <div class="well" id="cfp-errors">
+        <p>
+            <b>Lightning Talks</b> are short (5minutes) talks presented all together.
```
marksteward, line 10: Please add a space, so "5 minutes" (there's 4 of these)

apps/cfp/views.py
```diff
@@ -372,10 +461,11 @@ def complete() -> ResponseReturnValue:
 @login_required
 def proposals() -> ResponseReturnValue:
     proposals = current_user.proposals
-    if not proposals:
+    lightning_talks = current_user.lightning_talks
+    if not proposals or not lightning_talks:
```
marksteward, line 465: This should be `and` instead of `or`

marksteward (changes requested):

marksteward:
Note that while this PR includes code to show lightning talks on the /cfp page when they're available, we've set CFP_CLOSED. So people should instead be directed to https://www.emfcamp.org/cfp/lightning-talk once LIGHTING_TALKS has been enabled.

Or we need a bypass of CFP_CLOSED when LIGHTING_TALKS is set, maybe a redirect.

PR #2332, opened 2026-07-15
Make shift page viewable by more than just admins

ormiret:
Still needs to be logged in as a volunteer for the sign up button. Can probably extend this to point at sign up to volunteer then sign up to the shift but that is a tomorrow problem.


marksteward (approved):

marksteward:
lgtm, does @jellybob want a look?

PR #2340, opened 2026-07-16
Add volunteer shift subscription and iCal download links

silverphish-io:
Added a section for subscribing to shifts and downloading iCal file. Uses the existing functions and routes in apps/volunteer/schedule.py and just adds a href in an easy to access space.

For example, going to www.emfcamp.org/volunteer/schedule.ics will generate the current users volunteer shifts in an ics file. When viewed, the contents of the ics file match the shifts signed up for the user.

This is technically untested (as Im doing this on the drive to EMF), so I can't validate it. However *theoretically* it should be fine 🤞


jellybob:
❤️ Thank you! I've been meaning to get round to this all week!

PR #2344, opened 2026-07-17
Corrected nullsector opening times

6tee4:


marksteward:
I'm struggling to parse this. Are you saying NullSector is making noise throughout the event from this evening? We usually consider 5am our day changeover time, so 00:00-24:00 might be clearer as "all day", but we have an amplified noise curfew at like 2am.

marksteward:
Please also update the text in the paragraph before.

PR #2347, opened 2026-07-17
Show drop-in things on now and next until they finish

Jonty:


SamLR (approved):
I think that makes sense to me

PR #2352, opened 2026-07-18
Allow emailing people who have unclaimed products

lukegb:
We don't currently support emailing people who have unclaimed products from any product group, and we can't include information about their specific unclaimed products, but this is a good stopgap, at least.

### DIFF
--- a/apps/admin/email.py
+++ b/apps/admin/email.py
@@ -1,5 +1,3 @@
-from typing import Literal
-
 from flask import flash, redirect, render_template, url_for
 from flask.typing import ResponseReturnValue
 from sqlalchemy import select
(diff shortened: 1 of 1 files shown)


apps/admin/email.py
```diff
@@ -78,13 +105,17 @@ def get_email_reason(dest: str) -> str:
         return (
             f"You're receiving this email because you have a ticket or a talk/workshop accepted for {event}."
         )
+    if dest.startswith("unclaimed:"):
+        product = product_from_dest(dest)
+        return f"You're receiving this email because you have purchased {product.display_name} but have not yet redeemed your purchase."
```
marksteward, line 110: We should probably find a way to do this that creates nice wording, will make a note.

marksteward (approved):

PR #2364, opened 2026-07-25
Promote video streaming to front page properly

dxf:
Given that the talks are up in record time (<3 to the team + volunteers who made it happen), figured it was probably about time to update the page.


dxf:
many thanks :)

PR #2369, opened 2026-07-26
Fix SQL exception when adding/removing favourites

### LINKED ISSUE #2357: Catch SQL exception when adding/removing favourites
e.g.
```
app-1  | sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "pk_favourite_schedule_item"
app-1  | DETAIL:  Key (user_id, schedule_item_id)=(422, 124) already exists.
app-1  |
app-1  | [SQL: INSERT INTO favourite_schedule_item (user_id, schedule_item_id) VALUES (%(user_id)s, %(schedule_item_id)s)]
app-1  | [parameters: {'user_id': 422, 'schedule_item_id': 124}]
```

rixx:
closes #2357

Added two helper methods on `User`, one to set a favourite and one to toggle (for convenience). The setter upserts when adding a favourite and safely deletes when dropping it, so we should be good on race conditions (`DELETE ... WHERE ...` is idempotent, so I don't think we need the row lock mentioned in the issue).

I added an autouse fixture to `conftest.py` to clear the logged in user set by `flask-login`, as it was leaking between tests, as `app` is a module scoped fixture that keeps its context (including the user) between tests. I'm more of a Django dev and don't know what current practice is in Flask; if you want me to drop this part and work around it in my tests, say the word.

### DIFF
--- a/apps/api/schedule.py
+++ b/apps/api/schedule.py
@@ -89,18 +89,13 @@ class FavouriteScheduleItem(Resource):
             abort(401)
 
         schedule_item = get_or_404(db, ScheduleItem, schedule_item_id)
-        current_state = schedule_item in current_user.favourites
 
         data = request.get_json()
         if data.get("state") is not None:
             new_state = bool(data["state"])
+            current_user.set_favourite(schedule_item, new_state)
         else:
-            new_state = not current_state
-
-        if new_state and not current_state:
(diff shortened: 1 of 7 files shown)


marksteward:
Looks great, thanks for this!

PR #2370, opened 2026-07-27
Added all of the 2026 installations

jonathanhogg:
Processed automagically from the Art & Installations grist.

Let us not do it this way again.


Jonty:
These seem to be missing the name of the person who created them, which I suspect you want on there...

jonathanhogg:
> These seem to be missing the name of the person who created them, which I suspect you want on there...

Each artist was asked to put a credit into the description if they wanted one – some did and some didn't. Sadly the whole "finalise" mechanism for installations went away in the CfP refactor and so we had no easy way of getting a publishable name and pronouns.

PR #2374, opened 2026-07-29
Only try to push images in upstream repo

rixx:
Since opening my fork, every couple of days, I got notifications that the build for `rixx/Website` failed because it could not push the `emfcamp/Website` docker image. If I got the weird GitHub actions stringy bash/yaml/typing right, this should only try to push from the upstream repo, but still build on forks to see if the build works.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -29,11 +29,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


marksteward:
Probably better in the long-term to make the tags into variables, and skip the jobs if they're not set, but this seems reasonable for now.

PR #2377, opened 2026-07-31
Add a task for refunding/partrefunding keebspansions

lukegb:
This ended up a bit more complicated to deal with edgecases that I'm not convinced exist in practice but may be useful to handle in future years:

- Payments for a single, 'part-collected' keebspansion -> payment marked part-refunded, purchase changes price/price tier/product
- Payments for a single, uncollected keebspansion -> payment marked refunded, purchase marked refunded
- Payments that include those things but also other stuff -> payment marked part-refunded, purchase handled as above
- Payments that include a combination of part-collected and uncollected keebspansions -> payment marked part-refunded, purchases handled as above
- Users with keebspansion payments in multiple currencies, with some combination of the above -> does mostly the right thing as above

### DIFF
--- a/apps/tickets/tasks.py
+++ b/apps/tickets/tasks.py
@@ -535,7 +535,8 @@ def emf2026_refund_keebdecks_emailtest() -> None:
 @click.option(
     "--dry-run", is_flag=True, help="If set, don't actually refund/change the database, just simulate it."
 )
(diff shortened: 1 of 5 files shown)


lukegb:
I haven't actually tested this at all, so other than sprinkling `assert` everywhere to make mypy happy caveat emptor; this of all things I think I should write some actual tests for.

russss (approved):

PR #2379, opened 2026-08-05
Always show signed up shifts on the schedule

### LINKED ISSUE #1944: Always show signed up shifts on the schedule
A few reasons for this:

1. Finding which shifts they're on is one of the more likely reasons for someone being on the schedule page.
2. It makes it easier to spot when a shift would clash.
3. Its really confusing when people click sign up on a shift and it vanishes because adding themselves to it put it over the threshold for one of their other filters.

1Cazarez:
Fixes #1944

Problem: The show_signed_up_only filter ("Show my shifts") is displayed to all users on the volunteer schedule page, including volunteer admins who use the view to audit shift coverage.

Expected Behavior: Regular volunteers should not see the "Show my shifts" filter since it adds little value outside the context of a volunteer admin auditing shift coverage. Admins should retain the filter for auditing purposes.

Current Behavior: The filter is shown to all users. For regular volunteers this causes confusion — shifts can vanish when other filters are active and the "Show my shifts" filter isn't on.

Fix: Wrapped the show_signed_up_only checkbox in {% if not is_admin %} in templates/volunteer/_schedule_filters.html, following the existing {% if is_admin %} pattern already used in the same template for other admin-only controls.

Testing: Added two unit tests in tests/volunteer/test_schedule_filters.py verifying the checkbox renders for non-admins and is hidden for admins. Full test suite passes (./run_tests — 151 passed, 17 skipped).

### DIFF
--- a/js/volunteer_schedule.js
+++ b/js/volunteer_schedule.js
@@ -81,9 +81,9 @@ function getNodeData(node) {
 }
 
 function shouldDisplayNode(node_data, filters, node) {
-  // If the signed up shifts filter is active, or we're not set to show shifts
-  // in the past then those take precedence over all others and we short
-  // circuit everything else.
+  // Always show shifts the user is signed up for, and don't hide shifts
+  // in the past if the show_past filter is active. These take precedence
+  // over all other filters.
   if (node_data["notice"]) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


jellybob:
That's intentional behaviour as the same view is used by the volunteer team to audit where shifts are open, and to show other people, and in that case showing the user's own shifts isn't hugely useful. I'd consider a patch which instead hides that filter from users who aren't volunteer admins since it does seem that functionality is less useful outside the context of a volunteer admin.

1Cazarez:
Thanks for the context! I've updated the PR to instead hide the 'Show my shifts' filter from non-admin users, following the same {% if is_admin %} pattern already used in that template. Let me know if this looks like the right approach.

PR #2380, opened 2026-08-05
Fix post-event schedule appearing as empty

### LINKED ISSUE #2368: post-event schedule view is not very user-friendly
Image

Image

This should probably default to everything being visible after the event has ended.

rixx:
closes #2368

### DIFF
--- a/js/schedule/ScheduleData.jsx
+++ b/js/schedule/ScheduleData.jsx
@@ -16,21 +16,26 @@ class ScheduleData {
     this.ageRanges = [];
     this.ageRangesSeen = new Set();
 
-    this.allFinished = true;
+    let scheduleItems = this.rawSchedule.map((row) =>
+      this.parseScheduleItem(row),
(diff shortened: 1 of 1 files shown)


marksteward:
This breaks filtering on finished occurrences. I think it's better just to default includeFinished to true once the event's over. Or ignore it, but again only once the event is over.

rixx:
Hmm, my test setup must be wonky, it didn't break filtering for me. It took me a bit to figure out how to make a schedule appear in testing, so I must have gotten something wrong along the way. Sorry for the noise.

Updated to do it the way you suggested.

Slightly janky because now, the "Include finished events" checkbox does nothing after the event (unchecking it doesn't give you an empty page, which would be the alternative, but not really any better). Do you want to disable or hide the filter option once the event is over?

rixx:
Schedule still appears empty, but probably not super relevant, closing.

